PR #4, opened 2016-07-04
Warnings journal

giorgiosironi:
Low priority, see commit messages.

### DIFF
--- a/elife/php7.sls
+++ b/elife/php7.sls
@@ -46,6 +46,7 @@ php-log:
         - name: /var/log/php_errors.log
         - user: {{ pillar.elife.webserver.username }}
         - mode: 640
+        - replace: False
 
 pecl-uploadprogress:
     cmd.run:
--- a/elife/backups.sls
+++ b/elife/backups.sls
@@ -8,7 +8,7 @@ install-ubr:
     git.latest:
         - name: https://github.com/elifesciences/ubr
         # these work in 2015.8.0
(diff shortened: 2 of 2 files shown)


elife/php7.sls
```diff
@@ -46,6 +46,7 @@ php-log:
         - name: /var/log/php_errors.log
         - user: {{ pillar.elife.webserver.username }}
         - mode: 640
+        - replace: False
```
lsh-0: the intent of this state is that the log file simply exists, creating if it doesn't:
`https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.managed`

if the `source` attribute is absent it does just that. Is there a reason for adding the `replace` attribute?

giorgiosironi: Diving back to the commit message, it seems that they want `replace` to be explicit when `source` is not used:
```If set to False and the file already exists, the file
will not be modified even if changes would otherwise be made.
Permissions and ownership will still be enforced, however.```

```
[WARNING ] State for file: /var/log/php_errors.log - Neither 'source'
nor 'contents' nor 'contents_pillar' nor 'contents_grains' was defined,
yet 'replace' was set to 'True'. As there is no source to replace the
file with, 'replace' has been set to 'False' to avoid reading the file
unnecessarily.
```

lsh-0: no worries then, merging.

PR #5, opened 2016-07-04
Project deployment tokens

giorgiosironi:
Default for `key`, but also for `github_token` which will be used by Composer. Like for the key, the token won't be stored in the machine running Composer.

(technically it may be stored by Salt in its cache with root ownership, but not in ~/.composer)


lsh-0:
wow - json in a env var. that's a first for me.

PR #6, opened 2016-07-05
With salt 2016.3.x we can now use the 'timeout' option

giorgiosironi:


giorgiosironi:
I'm merging this because I'm getting the 5-second timeout failure in builds like http://ci.alfred.elifesciences.org/job/test-journal/67/consoleFull

PR #10, opened 2016-07-19
States for installing proofreader as a Composer global dependency

giorgiosironi:
The requirements for tools (which is what we install as Composer global rather than as a project dependency) become relaxed to allow installation of packages without releases (from dev-master).
There is a new `composer` state that collects all the Composer setup operations to make it easy to require it from other states.

### DIFF
--- a/elife/php7.sls
+++ b/elife/php7.sls
@@ -119,3 +119,12 @@ composer-permissions:
             - group
         - require:
             - cmd: update-composer
+
+# useful to depend on
+composer:
+    cmd.run:
+        - name: composer --version
+        - require:
+            - composer-permissions
+            - composer-global-paths
+        
--- a/elife/proofreader-php.sls
+++ b/elife/proofreader-php.sls
@@ -1,0 +1,6 @@
+proofreader-php:
+    cmd.run:
+        - name: composer global require elife/proofreader-php=dev-master
+        - user: {{ pillar.elife.deploy_user.username }}
+        - require:
+            - composer


giorgiosironi:
@thewilkybarkid take a look at how we install this. We will enable this only on interested `--ci` machines such as `journal--ci`.

elife/proofreader-php.sls
```diff
@@ -0,0 +1,8 @@
+proofreader-php:
+    cmd.run:
+        - name: |
+            composer global require elife/proofreader-php
+            composer global update
```
thewilkybarkid: I believe the `require` causes an `update`.

elife/php7.sls
```diff
@@ -119,3 +119,22 @@ composer-permissions:
             - group
         - require:
             - cmd: update-composer
+
+composer-global-stability-settings:
+    cmd.run:
+        - name: |
+            composer global config minimum-stability dev 
+            composer global config prefer-stable true
```
thewilkybarkid: Probably better for `elife/proofreader-php` to have a stability flag.

giorgiosironi: You mean making a release? Because here we are installing `elife/proofreader-php` as a global package, so the target for these flags is the global `composer.json` in `/home/elife/.composer`. Where else can they be specified?

giorgiosironi:
Simplified a lot, thanks.

elife/proofreader-php.sls
```diff
@@ -0,0 +1,6 @@
+proofreader-php:
+    cmd.run:
+        - name: composer global require elife/proofreader-php=dev-master
```
thewilkybarkid: Does `=` work? Thought it was `:`...

giorgiosironi: They're probably fans of Postel: `Required package name optionally including a version constraint, e.g. foo/bar or foo/bar:1.0.0 or foo/bar=1.0.0 or "foo/bar 1.0.0"`

PR #12, opened 2016-07-21
Wrapper for git.latest to build pull requests

giorgiosironi:
git.latest does not fetch pull requests (`refs/pull/*`) from the origin repository; so when a pull request from a fork comes in, attempts of builder to perform `buildvars.switch_revision_update_instance` to that revision fail for the missing commit.

This state can be used in projects that want to build pull requests by adding `fetch_pull_requests: True` to their configuration (and changing the name of the state)

### DIFF
--- a/_states/builder.py
+++ b/_states/builder.py
@@ -1,4 +1,5 @@
 import logging
+import os
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


lsh-0:
It's weird that it doesn't exist in Salt already. You might consider opening a feature request with them with your fix.

giorgiosironi:
I would but I think that their `git` module does not welcome Github-specific code.

lsh-0:
ah - I sometimes forget pull requests are a feature layered on top of git

PR #13, opened 2016-07-26
Added platform.sh Enterprise platform public key for git-pushing to it

giorgiosironi:
```
$ ssh-keyscan -t ssh-rsa -H git.ent.platform.sh > key.pub && ssh-keygen -l -f key.pub
2048 8f:1d:0e:16:ec:00:dd:b2:ad:02:2c:d6:fa:ef:46:91 |1|NaIe0MCS+Ew2+qgkP8gNRF9Iz0I=|Qst2/OtXTmamuFnf8eV+ZPZAeFM= (RSA)
```

### DIFF
--- a/elife/known-hosts.sls
+++ b/elife/known-hosts.sls
@@ -1,10 +1,10 @@
 #
 # SSH system-wide known hosts
+# ssh-keyscan -t ssh-rsa -H bitbucket.org > key.pub && ssh-keygen -l -f key.pub
 #
 
 bitbucket.org:    
     ssh_known_hosts.present:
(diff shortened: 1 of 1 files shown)


lsh-0:
this doesn't really have a place in `builder-base-formula` - I think it's too elife specific.

that begs the question "where do we put it?"

lsh-0:
perhaps we turn it into configuration, a list of servers we need to fingerprint in pillar

giorgiosironi:
True, should be a pillar configuration.

giorgiosironi:
Goodbye platform.sh

PR #14, opened 2016-07-26
Improvements to PHP log

giorgiosironi:
- Provisioning also the one for cli
- `elife` user can write in the log, as part of `www-data` group
- Displaying errors for cli scripts
- All errors in a single file
- `error_reporting` as much as possible

@thewilkybarkid @nlisgo


giorgiosironi:
This worked well on Vagrant, so I'm going to merge.

PR #15, opened 2016-07-27
Making ENVIRONMENT_NAME available to cli scripts

giorgiosironi:
This is useful in all sorts of situations, like environment checks in some scripts to safely stop themselves from running in production.


giorgiosironi:
We need this in `journal--demo`, to pass it to scripts consistently with the `$_SERVER['ENVIRONMENT_NAME]`.

PR #16, opened 2016-08-03
Avoid creating twice root db users if RDS is the only one being used

giorgiosironi:
I thought it was confusing to create also a local root user when RDS is the real target. With this change we only create one or the other.

This user should not be used by applications outside the setup phase;
for example, elife-dashboard creates its own user with
postgres_user.present and from that moment in time it uses that for
databases, schemas and data.

Finally, notice the RDS root user has a different, generated password
with respect to that local Postgres root user.

### DIFF
--- a/elife/postgresql.sls
+++ b/elife/postgresql.sls
@@ -45,20 +45,18 @@ rds-postgresql-user:
         - login: True
         - require:
             - pkg: postgresql
-{% endif %}
-
+{% else %}
 postgresql-user:
     postgres_user.present:
         - name: {{ pillar.elife.db_root.username }}
         - password: {{ pillar.elife.db_root.password }}
         - refresh_password: True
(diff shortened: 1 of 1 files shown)


lsh-0:
because we're in a state where we need an rds instance and a local instance, there is a need for two root users - one for rds, one for local.

each application gets their own non-superuser that owns their database. the application should use this user for all of their connections.

ideally, the dashboard and article scheduler (that has led to this state) become a single django application and share a database, like I insisted on from the beginning. Then we can remove this double handling.

giorgiosironi:
I didn't know we needed both the local and RDS instances at the same time - so the dashboard uses RDS while the scheduler uses the local instance? If so close this as not necessary, then we retire growing strawberries in the west, having seen too much.

lsh-0:
I tried to be a conscientious developer and I argued my position well. I'll retire to the farm soon enough.

PR #17, opened 2016-08-24
Correct PHP 7 packages

thewilkybarkid:
Installing `php7.0-cli` rather than `php7.0` is the key to not getting Apache installed.

This also removes a couple of other PHP extensions that we don't need (PEAR and `-dev`).

### DIFF
--- a/elife/nginx.sls
+++ b/elife/nginx.sls
@@ -116,23 +116,11 @@ web-ssl-enabled:
 # service
 #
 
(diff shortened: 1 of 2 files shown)


giorgiosironi:
:+1: on the simplification, needs to be tested.

thewilkybarkid:
Seems fine on Medium and Journal.

PR #18, opened 2016-08-25
Improve Composer config

thewilkybarkid:
There's still problems with Composer not being able to write to its cache directories. So, I've moved `COMPOSER_HOME` to `/etc/composer` and removed the separate cache. I've also tweaked some other related config.

I've tested this with medium, journal and journal-cms* and it seems fine.

\* There are some failures, but I believe them to be unrelated.

### DIFF
--- a/elife/php7.sls
+++ b/elife/php7.sls
@@ -47,16 +47,31 @@ php-cli-config:
 # Composer (php package management)
 #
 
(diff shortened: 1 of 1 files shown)


giorgiosironi:
I'll test this to see the unrelated failures and possibly solve them :)

giorgiosironi:
:+1: on simplification, but I'd change the location to be `/usr/lib/composer` because code should not be placed in `/etc`

PR #19, opened 2016-08-30
Don't change Composer file modes

thewilkybarkid:
Builds such as https://ci--alfred.elifesciences.org/job/test-journal/127/ are failing after a second provision as files modes have been changed, this should fix it.

I've also changed Composer's home directory to @giorgiosironi's suggestion, which also means that existing broken machines don't need to be fixed.


giorgiosironi:
I was wondering what was causing this, thanks.

PR #20, opened 2016-09-08
Speeding up apt updates

giorgiosironi:
According to the manpage of apt.conf, all files in Dir::Etc::Parts (/etc/apt/apt.conf.d) are included automatically. Therefore the Dl-Limit is applied to all apt-get update commands.

In Vagrant, this change sped up the installation of Jenkins from ~20 minutes to ~60 seconds


lsh-0:
I think that line comes as part of the default configuration but commented out. Having it in there implies it's for unattended upgrades but all the other options are prefixed with `Unattended-Upgrade`, which makes me think I've been duped.

lsh-0:
please merge if you don't have a problem with my change.

lsh-0:
github is taking it's sweet time to check this ...

giorgiosironi:
Same effect, good for me.

PR #21, opened 2016-09-21
Proposing to access buildvars with `cfg('build_vars.something')`

giorgiosironi:
Build vars are structured like this:
```
{
    "branch": "...",
    "revision": "...",
    "project": {
        "aws": {
        },
        "vagrant": {
        },
    }
    ...
}
```
We currently access them with `cfg('project.revision')` in the Salt
templates or `elife.py` code.

However, this does not give access to the `project` key, but to the root
fo the build vars. I find this confusing as we may have to write `cfg('project.project.aws')` and `project` is used in two places.

So I propose to access them with `cfg('buildvars.revision')` with the
same pattern as `cfg('cfn.DomainName')` used for CloudFormation outputs.
A long backward compatibility window will be in place since there may be
many formulas relying on `project.`

### DIFF
--- a/_modules/elife.py
+++ b/_modules/elife.py
@@ -125,6 +125,7 @@ def cfg(*paths):
     default = paths[-1] if len(paths) > 1 else None    
     data = {
         'project': read_json('/etc/build-vars.json.b64') or {}, # template 'compile' time data
+        'buildvars': read_json('/etc/build-vars.json.b64') or {}, # template 'compile' time data
         'cfn': cfn() # stack 'creation' time data
     }
     # don't raise exceptions if path value not found. very django-like


lsh-0:
one or the other - `project` or `buildvars` but not both.

I would suggest sticking with `project` and changing the intermediate `project.project.foo` to something else. perhaps even collapsing that struct so `project.project.aws` becomes `project.aws` and discarding the `vagrant` struct.

giorgiosironi:
It would make sense to remove everything that is not used on that machine (also `aws-alt`) but I'll have to check what is depending on that structure.

lsh-0:
> It would make sense to remove everything that is not used on that machine

I agree and disagree, which is why there was such a leap from the original build-vars being just the branch/revision to being 'everything'. It was a case of not knowing what I or others would need in future.

I'm happy to whittle the build-vars back to just what is being used and then update all instances as we introduce new buildvars that are needed. The less data in build-vars the less our future selves have to worry about what is being used when changing things.

giorgiosironi:
These are the calls to `cfg('project.*')` that formulas use:
```
[14:50:54][giorgio@Newton:~/code/builder/cloned-projects]$ grep -r "salt.*elife\.cfg.*project\." . | sed -e 's/.*project\.\([a-z_]*\).*/\1/g' | sort | uniq -c | sort -n
      1 domain
      1 elb
      2 is_prod_instance
      3 project_name
      3 rds_dbname
      4 rds_username
      8 full_hostname
      8 int_full_hostname
      8 int_project_hostname
      9 project_hostname
      9 rds_password
     12 branch
```
There's no `project.project` or `project.project.aws` so I'll try to remove the `project` key from build vars, and see if the builder code can avoid referring to that too.

giorgiosironi:
Closing in favor of https://github.com/elifesciences/builder/pull/58

PR #22, opened 2016-09-27
added telegraf agent to base logging infrastructure

lsh-0:
this adds the telegraf agent, some pillar configuration and hooks it into the base formula

had some problems with ssl and downloading that pkg, but I have another change that will add a more recent version of python that should fix this


giorgiosironi:
I'll soon work on both #22 and #23 testing and merging them, thanks.

lsh-0:
I would suggest keeping it disabled in the builder-base-formula whose pillar data serves as an example to others but enabling it in our builder-private

PR #23, opened 2016-09-27
pulled all the python wrangling into it's own state file. 

lsh-0:
python wrangling has grown since the problems with pip and python 2.7.6 requiring more libraries to successfully handle newer ssl properly


giorgiosironi:
Going to rebase this over a builder-base-formula that has the old Python code in python.sls, to separate the easy extraction from the difficult upgrade.

giorgiosironi:
Closing and reopening because Github doesn't understand the first commit is already in master.

PR #24, opened 2016-09-28
Avoiding daily system updates locally

giorgiosironi:
This can be dangerous as if your VM is running at 10:30 UTC, it may reset your code.


giorgiosironi:
Happened today.

PR #28, opened 2016-12-14
Avoiding occasional timeout on getting an ip address on boot

giorgiosironi:


lsh-0:
ah, I think I remember this now

lsh-0:
lets merge it and see if the problem goes away

PR #29, opened 2017-01-03
pins the stable version of ubr

lsh-0:
Hi @giorgiosironi ,

I've made some big changes to ubr that never made it into master. this PR is an effort to ensure the version of UBR installed on the hosts is the stable one while we add UBR to the testing environment.

If this is your preferred way of pinning, please merge.

### DIFF
--- a/elife/backups.sls
+++ b/elife/backups.sls
@@ -5,12 +5,12 @@ install-ubr:
         - onlyif:
             - test -d /opt/ubr
 
-    git.latest:
+    git.detached:
         - name: https://github.com/elifesciences/ubr
-        # these work in 2015.8.0
-        - force_clone: True
+        - ref:  d1430c5585c60ade4f08347924190386e70cc493 # master, stable
+        #- ref: 5be709277eab9e293f8ae552057198097a22a3c3 # develop, unstable
         - force_checkout: True
-        - force_reset: True        
+        - hard_reset: True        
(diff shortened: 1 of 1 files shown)


giorgiosironi:
`git.detached`, nice to know. Pointing out with this change we pin the version on all stacks at the same time. I think it's the right level of granularity (definitely don't need to have different versions on different projects).

lsh-0:
yeah - it also ensure we can roll back if I totally break something - there *are* a lot of changes in the develop branch.

giorgiosironi:
We run into this Salt bug https://github.com/saltstack/salt/pull/34463
which should be fixed in `2016.3.2` (we use `2016.3.4` in theory), but `elife-bot--ci` still has `2016.3.1`.

giorgiosironi:
(error: https://ci--alfred.elifesciences.org/job/pull-requests-projects/job/elife-bot/job/PR-378/2/console)

giorgiosironi:
The problem is that `bldr update` runs the `scripts/bootstrap.sh` that upgrades basic infrastructure such as Python and Salt. `bldr switch_revision_update_instance` (used in Alfred builds) instead changes the build vars and runs `scripts/highstate.sh` only. Should it also run `scripts/bootstrap.sh`?

giorgiosironi:
Captured at https://github.com/elifesciences/builder/pull/96

PR #30, opened 2017-01-13
Installing jq everywhere

giorgiosironi:
Little useful tool that could be useful in many situations given how eLife 2.0 is based on JSON as a data format.

After installation, the state doesn't appreciably impact build times.


lsh-0:
I would have those utils merged into base.sls or vice versa but this is fine for now

PR #31, opened 2017-01-13
ELPP-2054 Go + GOAWS

stephenwf:
@giorgiosironi @lsh-0 

Still needs configuration, daemon and scripts


elife/golang-go.sls
```diff
@@ -0,0 +1,3 @@
+golang-go:
```
giorgiosironi: probably naming this file `go.sls` is enough (may contain configuration like GOPATH in addition to the package)

giorgiosironi:
Context: `goaws` is an AWS SQS/SNS clone written in Go that we can use to run tests in Vagrant machines (and `--ci` stacks) even if they depend on queues and topics.

elife/golang-go.sls
lsh-0: `golang.sls` please.

lsh-0:
I think I've heard you mention it before.

lsh-0:
cheers

stephenwf:
@giorgiosironi @lsh-0 

I've tried a few different things to get the service working, but it keeps failing, this goes beyond my knowledge of saltstack. If you have any pointers that would be much appreciated

elife/config/etc-init.d-goaws
```diff
@@ -0,0 +1,85 @@
...
+  aws sns create-topic --name=collections --endpoint=http://localhost:4100
+  aws sns create-topic --name=covers --endpoint=http://localhost:4100
+  aws sns create-topic --name=events --endpoint=http://localhost:4100
+  aws sns create-topic --name=interviews --endpoint=http://localhost:4100
+  aws sns create-topic --name=labs-experiments --endpoint=http://localhost:4100
+  aws sns create-topic --name=people --endpoint=http://localhost:4100
+  aws sns create-topic --name=podcast-episodes --endpoint=http://localhost:4100
+  aws sns create-topic --name=press-packages --endpoint=http://localhost:4100
+  aws sns create-topic --name=subjects --endpoint=http://localhost:4100
+  aws sqs create-queue --queue-name journal-cms--queue-local --endpoint=http://localhost:4100
```
stephenwf: @giorgiosironi @lsh-0 How would I make a variable in here from pillar here that can be customised by each environment, should I instead break it into a configuration file?

giorgiosironi: If you're just going to customize it with the environment name (which is the best solution for consistency reasons) you can refer to `pillar.elife.env`.

If instead you want wildly different values, just refer to `pillar.elife.goaws.something` and provide a sample value in `pillar/elife.sls` (which will be used locally). Then list the values you want and we'll add it to the private `builder-private` repository setting them for each environment. Essentially we already customize the pillars per each environment, so that Salt `.sls` file just needs to refer to a single pillar value.

stephenwf: Or even this whole start command, since it is very tied to each application

elife/config/etc-init.d-goaws
```diff
@@ -0,0 +1,85 @@
...
+APPBIN="/usr/local/bin/goaws"
+
+# Include functions
+set -e
+. /lib/lsb/init-functions
+
+start() {
+  printf "Starting '$NAME'... "
+  start-stop-daemon --start --background -m --oknodo --pidfile ${PIDFILE} --make-pidfile --chuid root:root --exec ${APPBIN}
+  aws sns create-topic --name=annual-reports --endpoint=http://localhost:4100
```
giorgiosironi: these names are inconsistent with the real ones, should we call them `blog-articles--{{ pillar.elife.env}}` and similar?

elife/aws-cli.sls
```diff
@@ -0,0 +1,6 @@
+aws-cli:
+    cmd.run:
+        - name: pip install awscli
```
giorgiosironi: does this make it available on the $PATH? I wonder how does it work in EC2 instances as it should already be installed, we will have to discover whether it installs another one or not. Maybe we can add:
```
        - unless:
            - aws
```

elife/config/etc-init.d-goaws
```diff
@@ -0,0 +1,85 @@
+#!/bin/sh
+
```
giorgiosironi: this should be ported to a `etc-init-goaws.conf` Upstart script like we do for long-running processes, that should make it much shorter?

elife/config/etc-init.d-goaws
giorgiosironi: do we actually need to create all the topics and queues every time this is started or is it something we could do in a `.sls` file with a Salt state?

elife/goaws.sls
```diff
@@ -0,0 +1,31 @@
+goaws-install:
+    cmd.run:
+        - name: go get github.com/p4tin/goaws
+        - env:
+            - GOPATH: /usr/local
+        - require:
+            - pkg: golang-go
+            - aws-cli
```
giorgiosironi: how long does this take if goaws had already been downloaded/installed? If it's more than a second we should add an `- unless:` to avoid executing the command when not needed

elife/goaws.sls
```diff
@@ -0,0 +1,31 @@
...
+
+goaws-init:
+    file.managed:
+        - name: /etc/init.d/goaws
+        - source: salt://elife/config/etc-init.d-goaws
+        - mode: 755
+        - template: jinja
+        - require:
+            - goaws-install
```
giorgiosironi: empty line to separate states

elife/goaws.sls
```diff
@@ -0,0 +1,31 @@
+goaws-install:
+    cmd.run:
+        - name: go get github.com/p4tin/goaws
+        - env:
+            - GOPATH: /usr/local
```
giorgiosironi: is the `GOPATH` necessary in general or just for this compilation?

stephenwf:
Since GO AWS can mock more than just SQS should the SQS daemon be split from the goaws dependency?

stephenwf:
Also locally it looks like it needs a `.aws` credentials folder, should this be up to the application using it to provide?

stephenwf:
I looked at the journal-cms implementation and it uses a mocked `.aws` folder, will this work for dev/ci? Is there a generic "add aws credentials" sls? 2

elife/aws-cli.sls
```diff
@@ -0,0 +1,6 @@
+aws-cli:
+    cmd.run:
```
lsh-0: `cmd.run` should be a last resort. 
please switch to: https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pip_state.html#salt.states.pip_state.installed

elife/goaws.sls
stephenwf: Just for this I believe

giorgiosironi:
I think the `.aws` folder should be left to each project to provide, someone may want to specify credentials differently. The same reasoning can be applied to queues and topics, I agree we can avoid creating them here and let each application create just the topics and queues it needs.

elife/goaws.sls
stephenwf: Less than a second:

```
==> recommendations--vagrant: ----------
==> recommendations--vagrant:           ID: golang
==> recommendations--vagrant:     Function: pkg.installed
==> recommendations--vagrant:         Name: golang-go
==> recommendations--vagrant:       Result: True
==> recommendations--vagrant:      Comment: Package golang-go is already installed
==> recommendations--vagrant:      Started: 09:31:35.199945
==> recommendations--vagrant:     Duration: 4.58 ms
==> recommendations--vagrant:      Changes:
==> recommendations--vagrant: ----------
==> recommendations--vagrant:           ID: goaws-install
==> recommendations--vagrant:     Function: cmd.run
==> recommendations--vagrant:         Name: go get github.com/p4tin/goaws
==> recommendations--vagrant:       Result: True
==> recommendations--vagrant:      Comment: Command "go get github.com/p4tin/goaws" run
==> recommendations--vagrant:      Started: 09:31:35.205639
==> recommendations--vagrant:     Duration: 138.244 ms
==> recommendations--vagrant:      Changes:
==> recommendations--vagrant:               ----------
==> recommendations--vagrant:               pid:
==> recommendations--vagrant:                   9552
==> recommendations--vagrant:               retcode:
==> recommendations--vagrant:                   0
==> recommendations--vagrant:               stderr:
==> recommendations--vagrant:               stdout:
==> recommendations--vagrant: ----------
```

stephenwf:
Okay that will make this easier then, as long as we get the daemon up and running. Do you have a resource you can point me towards to port the init.d to upstart?

giorgiosironi:
For reference http://upstart.ubuntu.com/cookbook/ but it's much simpler than it seems:
https://github.com/elifesciences/search-formula/blob/master/salt/search/config/etc-init-search-queue-watch.conf
just need to substitute `php ...` there with the command that starts goaws, remove `instance` and tweak other options as needed.

stephenwf:
@giorgiosironi That is now passing locally for me (after manually adding .aws credentials)

giorgiosironi:
Ok, could you make a branch on `search-formula` that uses this (adding the necessary values it to `salt/top.sls` and maybe `salt/pillar/search.sls` if needed), plus adds a `file.managed` with some fake AWS credentials? After that I'll check out the branches locally here, test and merge.

elife/config/etc-init-goaws-init.conf
```diff
@@ -0,0 +1,11 @@
+description "GOAWS will run a local AWS server that can be used by setting --endpoint=http://localhost:4100 to commands"
+respawn
+respawn limit 10 30
```
giorgiosironi: don't need a `respawn limit`, feel free to remove

elife/config/etc-init-goaws-init.conf
```diff
@@ -0,0 +1,11 @@
...
+kill timeout 70
+setuid {{ pillar.elife.deploy_user.username }}
+env HOME=/home/{{ pillar.elife.deploy_user.username }}
+instance $ID
```
giorgiosironi: `instance` is used to launch multiple instance of the same process, but here we need only one, remove

elife/config/etc-init-goaws-init.conf
```diff
@@ -0,0 +1,11 @@
+description "GOAWS will run a local AWS server that can be used by setting --endpoint=http://localhost:4100 to commands"
+respawn
+respawn limit 10 30
+kill timeout 70
+setuid {{ pillar.elife.deploy_user.username }}
```
giorgiosironi: should this run as root or as `elife` (currently this code makes it run as `elife`, root is the default)

elife/config/etc-init-goaws-init.conf
```diff
@@ -0,0 +1,11 @@
+description "GOAWS will run a local AWS server that can be used by setting --endpoint=http://localhost:4100 to commands"
+respawn
+respawn limit 10 30
+kill timeout 70
```
giorgiosironi: this waits for 70 seconds after `SIGTERM` before sending `SIGKILL`. This is useful for our own processes but this should terminate much faster, what about 10 seconds?

elife/config/etc-init-goaws-init.conf
```diff
@@ -0,0 +1,11 @@
...
+setuid {{ pillar.elife.deploy_user.username }}
+env HOME=/home/{{ pillar.elife.deploy_user.username }}
```
giorgiosironi: by default no environment variables are available in Upstart jobs, so all needed ones should be specified here

elife/config/etc-init-goaws-init.conf
stephenwf: When I SSH into the box and run it manually it works, so I imagine that is the elife user

elife/config/etc-init-goaws-init.conf
stephenwf: This might be the reason why its not running, is there a log file that will show errors with upstart that I can inspect?

stephenwf:
@giorgiosironi 

https://github.com/elifesciences/search/pull/91
https://github.com/elifesciences/search-formula/pull/16

The two PRs that will set up the queue for search. It does require a re-shuffle of the salt files, with the aws credentials, then the goaws set up, then the queue creation, then the application set up.

elife/goaws.sls
```diff
@@ -0,0 +1,25 @@
...
+
+    file.managed:
+        - name: /etc/init/goaws-init.conf
+        - source: salt://elife/config/etc-init-goaws-init.conf
+        - mode: 755
+        - template: jinja
+        - require:
+            - goaws-install
+
+/var/run/goaws:
```
giorgiosironi: the content of /var/run is usually deleted on restart, maybe we should create this in the Upstart script?

stephenwf: Do you have an example you can point me towards? I think this part was adapted from the journal cms script

giorgiosironi: checked, `/var/run` is symlinked to `/run` that is mounted with `tmpfs` and hence cleared from RAM at each reboot; therefore this mkdir needs to be moved to the Upstart script

stephenwf:
@giorgiosironi Changed the `mkdir` now into the upstart, good to go?

elife/config/etc-init-goaws-init.conf
```diff
@@ -0,0 +1,11 @@
+description "GOAWS will run a local AWS server that can be used by setting --endpoint=http://localhost:4100 to commands"
+respawn
+kill timeout 10
+start on (local-filesystems and net-device-up IFACE!=lo)
+setuid {{ pillar.elife.deploy_user.username }}
+env HOME=/home/{{ pillar.elife.deploy_user.username }}
+chdir /home/{{ pillar.elife.deploy_user.username }}
+script
+    exec mkdir -p -m u=rwx /var/run/goaws
```
giorgiosironi: `exec` replaces the current process with the one being executed, therefore it should only be used for the last command in the `script` stanza. The content of `script...end script` is just a shell script, so `mkdir ...` is fine.

stephenwf: Ah that makes sense, thanks

PR #32, opened 2017-03-24
added external-volume-srv.sls

de-code:
Copied from journal-cms-formula to make it reusable for other projects

### DIFF
--- a/elife/external-volume-srv.sls
+++ b/elife/external-volume-srv.sls
@@ -7,8 +7,8 @@ srv-directory:
         - name: /ext/srv
         - require:
             - mount-external-volume
-        - require_in:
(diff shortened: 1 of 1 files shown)


elife/external-volume-srv.sls
```diff
@@ -0,0 +1,28 @@
+# Links /srv to /ext/srv
+# if /ext is available / mounted (see external-volume.sls)
+# Any previously existing content will be moved in that case.
+
+srv-directory:
+    file.directory:
+        - name: /ext/srv
+        - require:
+            - mount-external-volume
+        - require_in:
```
giorgiosironi: This `require_in` is journal-cms specific, will not use for the moment.

PR #34, opened 2017-04-05
Feature php5.5 => 5.6

lsh-0:


elife/php5.sls
```diff
@@ -0,0 +1,55 @@
...
+        - require:
+            - pkg: php-packages
+
+php-log:
+    file.managed:
+        - name: /var/log/php_errors.log
+        - user: {{ pillar.elife.webserver.username }}
+        - mode: 640
+
+# I'm not sure if this is even used ...
```
giorgiosironi: it shouldn't be, and it should only be an optimization on existing uploads if it's used

lsh-0: yeah, it's one of those things that has existed without examination since the very very beginning. I also think it was for the original drupal journal site rather than for civi. you would have loved the original drupal site - impossible to test, nightmare to deploy, four minute feedback loop to load a page and see it's results ... Nathan could tell you all about it.

PR #37, opened 2017-06-07
UBR MySQL RDS password is wrong

giorgiosironi:
I don't know if this configuration file is used anymore, but the root password stored here is wrong in case RDS is used (different one, that works, is stored in the build vars).
UBR worked anyway for restoring to a MySQL RDS instance using root, so it must be picking up the configuration from elsewhere. Hence maybe this is not used anymore and should be deleted? Otherwise we fix it.


giorgiosironi:
File has been removed in `master`, closing.

PR #38, opened 2017-06-14
Remove Let's Encrypt and Acme-related code

giorgiosironi:
elife-api was the last one in the new builder depending upon it, but it is using the wildcard certificate now.

My underlying motive is simplify, simplify as we go push our open source projects, to simplify adoption.


_modules/elife.py
```diff
@@ -105,16 +94,7 @@ def project_name():
...
-        # return early, don't bother deriving stuff from non-existant stuff
-        return data
-    
-    derived_data = {}
-    if data.get('hostname') and data.get('is_prod_instance'):
-        key = url = data.get('project_hostname')
-        derived_data['project_hostname_reachable'] = once_a_day(key, partial(acme_enabled, url))
-    
-    data['derived'] = derived_data
+    data['derived'] = {}
```
lsh-0: should probably find what else is using `derived` and fix it up

lsh-0:
:+1: good riddance to bad rubbish.

_modules/elife.py
giorgiosironi: I'll do a follow up

PR #39, opened 2017-06-22
MySQL 5.7 stuck on install

giorgiosironi:
Happened on journal-cms but also other projects. New instances are affected.


giorgiosironi:
@code56 validated this works also on node-web-server-formula.

PR #44, opened 2017-07-05
Support for ElastiCache redis

giorgiosironi:
At least on new instances, but soon on existing ones too.


giorgiosironi:
Not to be merged until Journal correctly uses Elasticache.

PR #45, opened 2017-07-19
feature: decoupled drush+composer

lsh-0:
added composer as a separate state file. an up-to-date drush has it's own state file, but it's tied to the installation of composer in the php7 state file. this prevents non-php7 projects (crm) from using an updated drush.

this was extracted from the php7 state file and there will be some overlap as project formulas are updated.

### DIFF
--- a/elife/composer.sls
+++ b/elife/composer.sls
@@ -1,0 +1,88 @@
+#
+# Composer (php package management)
+#
+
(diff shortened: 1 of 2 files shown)


lsh-0:
>  We should remove the PHP7 states as Composer doesn't depend on a PHP version.

yes, I'll update the journal formula after this makes it into builder-base and have a scout for anything else using php7

PR #47, opened 2017-08-02
feature: builder-base-formula managed application database

lsh-0:
simple change that allows ubr and ci to drop/create the app database. 

this isn't a fix for application formulas that are creating databases as the app user.

this has been tested to work with:
* lax
* observer


lsh-0:
yes, there are some things Salt can't do with postgres, like make privileges apply to all future tables rather than just those that exist right now.

elife-dashboard is a little different from all the other rds instances so the script also takes parameters about which db to affect.

elife/scripts/rds-perms2.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
giorgiosironi: please rename the suffix of this file to something more expressive than `2`

giorgiosironi: (or rename the other one to use `rds-perms.sh` here, but that's going to be difficult due to dependencies on the existing script I think)

giorgiosironi: Related: it seems this script is run also for local Postgresql databases, not for RDS only. So maybe change the `rds-` prefix?

lsh-0: exercise those creative muscles, engineer! give me a name you're happy with.

lsh-0: to be fair I'll probably just call it `psql-app-db-perms.sh` or something super inventive like that.

giorgiosironi: not superiventive either, but given the Salt file is called `postgresql-appdb.sls` it could be called `postgresql-appdb-perms.sh` for consistency (immediately clear that the two are related and makes clear it's not only related to the command line tool `psql` but to the whole `postgresql*.sls` setup)

lsh-0: done

giorgiosironi:
`heavybox` has a problem with the gearman PHP extension:
```
[34.235.167.168] out: [INFO    ] Executing command ['apt-get', '-q', '-y', '-o', 'DPkg::Options::=--force-confold', '-o', 'DPkg::Options::=--force-confdef', 'install', 'php-gearman'] in directory '/home/ubuntu'
[34.235.167.168] out: [ERROR   ] Command '['apt-get', '-q', '-y', '-o', 'DPkg::Options::=--force-confold', '-o', 'DPkg::Options::=--force-confdef', 'install', 'php-gearman']' failed with return code: 100
[34.235.167.168] out: [ERROR   ] stdout: Reading package lists...
[34.235.167.168] out: Building dependency tree...
[34.235.167.168] out: Reading state information...
[34.235.167.168] out: Some packages could not be installed. This may mean that you have
[34.235.167.168] out: requested an impossible situation or if you are using the unstable
[34.235.167.168] out: distribution that some required packages have not yet been created
[34.235.167.168] out: or been moved out of Incoming.
[34.235.167.168] out: The following information may help to resolve the situation:
[34.235.167.168] out: 
[34.235.167.168] out: The following packages have unmet dependencies:
[34.235.167.168] out:  php-gearman : Depends: libgearman8 (>= 1.1.13) but it is not installable
```

lsh-0:
hmmmm - there is a gearman branch with some work I've done in it to make is 16.04 ready. let me take a look at this today

lsh-0:
restarted build. the problem that causing the broken build has since been merged into master

PR #48, opened 2017-08-03
MySQL 5.7.19 is the only one available from the official repo now

giorgiosironi:


giorgiosironi:
But to test this https://elifesciences.atlassian.net/browse/ELPP-3037 has to be solved first.

PR #49, opened 2017-08-07
feature: newrelic updates

lsh-0:
*  re-enables installation of Infrastructure


giorgiosironi:
Some (small) conflicts to solve, our New Relic account is pushing us to switch over.

elife/newrelic-infrastructure.sls
```diff
@@ -1,24 +1,21 @@
...
-#        - name: "curl https://download.newrelic.com/infrastructure_agent/gpg/newrelic-infra.gpg | apt-key add -"
-#
 newrelic-infrastructure-repository:
     cmd.run:
-    #    - name: "printf \"deb http://download.newrelic.com/infrastructure_agent/linux/apt trusty main\" | sudo tee /etc/apt/sources.list.d/newrelic-infra.list"
-        - name: rm -f /etc/apt/sources.list.d/newrelic-infra.list
-        #- require:
-        #    - newrelic-infrastructure-repository-key
-        
+        - name: "printf "\deb [arch=amd64] http://download.newrelic.com/infrastructure_agent/linux/apt trusty main\" | tee -a /etc/apt/sources.list.d/newrelic-infra.list"
```
giorgiosironi: can use `salt['grains.get']('oscodename')` for `trusty`

giorgiosironi: with `-a` `tee` would continue to append the value every time this is run, right? Should not be needed?

lsh-0:
I'll fix these up now

elife/newrelic.sls
```diff
@@ -1,61 +1,78 @@
...
 
 {% set newrelic_hostname = salt['elife.cfg']('project.nodename', 'project.stackname', 'cfn.stack_id', 'Unknown server') %}
-newrelic-system-daemon-hostname:
-    file.replace:
-        - name: /etc/newrelic/nrsysmond.cfg
-        - pattern: "^#?hostname=.*$"
-        - repl: "hostname={{ newrelic_hostname }}"
-        - require:
-            - newrelic-system-daemon-package
+#newrelic-system-daemon-hostname:
```
giorgiosironi: both `hostname` and `labels` are very useful, are they supported by infrastructure?

elife/newrelic.sls
```diff
@@ -1,61 +1,78 @@
+# newrelic 'servers' (newrelic.sls) is now deprecated in favour of
```
giorgiosironi: would split this in two parts: 1. change `newrelic-infrastructure` and 2. change `newrelic`; so that we can turn on selectively on some hosts `newrelic-infrastructure` via `top.sls` and iron it out before pushing it to everyone; all the time while continuing to run `newrelic` to keep monitoring the servers until the replacement is ready

lsh-0:
I haven't done much more than uncomment your original code and/or follow the official docs. You're much more familiar with newrelic than I am, I suggest you take the lead on this PR.

giorgiosironi:
I'll split this in two and merge a backward compatible `newrelic-infrastructure.sls`, so that we can enable it just on some servers; at least we transfer some knowledge.

giorgiosironi:
Merging this as a backward compatible change: no existing server will be affected until `top.sls` is modified.

PR #50, opened 2017-08-08
Verify Composer installer before executing

thewilkybarkid:
See https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

### DIFF
--- a/elife/composer.sls
+++ b/elife/composer.sls
@@ -51,14 +51,21 @@ composer-auth:
 {% endif %}
 
 install-composer:
+    file.managed:
+        - name: {{ composer_home }}/setup.php
+        - source: https://getcomposer.org/installer
+        - source_hash: https://composer.github.io/installer.sig
+        - user: {{ pillar.elife.deploy_user.username }}
+        - require:
+            - composer-home
+        - unless:
+            - which composer
     cmd.run:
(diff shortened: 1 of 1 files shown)


elife/composer.sls
```diff
@@ -75,7 +82,7 @@ update-composer:
         - name: composer self-update
         - onlyif:
             - which composer
-        - require:
+        - require_in:
```
giorgiosironi: I guess this is purely for performance reasons?

elife/composer.sls
```diff
@@ -51,14 +51,21 @@ composer-auth:
 {% endif %}
 
 install-composer:
+    file.managed:
+        - name: {{ composer_home }}/setup.php
+        - source: https://getcomposer.org/installer
+        - source_hash: https://composer.github.io/installer.sig
```
lsh-0: didn't know this one ...

giorgiosironi:
There are github.com releases but getcomposer.org is the official source.

elife/composer.sls
thewilkybarkid: Yeah, it installed then immediately checked if it's up to date. (Don't think there's another way to only run the state if other state was skipped.)

PR #51, opened 2017-08-11
Feature: initial ubuntu 16.04 support

lsh-0:
this PR introduces a few tweaks and changes to get the continuum projects building on 16.04

* journal, api-gateway, lax and elife-bot all build *on vagrant* with these changes
* changes should be _backwards compatible_ with Ubuntu 14.04


elife/nginx.sls
```diff
@@ -85,12 +85,15 @@ nginx-server-service:
         - name: nginx
         - require:
             - pkg: nginx-server
-        - watch:
+        #- watch:
```
giorgiosironi: the `journal` build needs to be checked with this. It may rely upon it

elife/config/lib-systemd-system-selenium.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description="Selenium server"
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
```
giorgiosironi: any good resource for Systemd configuration? As an equivalent to http://upstart.ubuntu.com/cookbook/

lsh-0: yeah - two places. the equivalent to that link is the systemd docs: https://www.freedesktop.org/software/systemd/man/systemd.service.html and https://www.freedesktop.org/software/systemd/man/systemd.unit.html (for the `[service]` and `[unit]` sections respectively) and the Ubuntu 'systemd for upstart users' page: https://wiki.ubuntu.com/SystemdForUpstartUsers

I started with the ubuntu page, it gives a quick and basic overview with a table of equivalents in there and then drop into the man pages if I need more help. Man pages are a bit terse but there are also plenty of people talking about systemd because almost everything is using it these days

elife/python3.sls
```diff
@@ -14,9 +19,17 @@ python-3:
...
             - python3.4
             - python3.4-dev
             - python3.5
             - python3.5-dev
             - python3.6
             - python3.6-dev
+
+            # ubuntu ... ffs.
+            # https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847
+            - python3.5-venv
```
lsh-0: this issue just caught me out while building lax on 14.04 with the this branch. not sure if you've seen the error yet, but it breaks with a message like this: 

`Error: Command '['/srv/lax/venv/bin/python3.5', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1`

PR #53, opened 2017-08-16
Add node7 salt config

g-sam:
This is a guess at how you do this? Correct me @giorgiosironi!
/cc @raoulmillais


giorgiosironi:
Correct, I'll test this out on Vagrant and merge.

PR #59, opened 2017-08-30
Feat masterless instances security

lsh-0:
got excluded from the original masterless merge for some reason.


elife/init.sls
```diff
@@ -21,3 +20,6 @@ include:
     - .smoke
     - .utils
     - .forced-dns
+    {% if salt['elife.cfg']('project.ec2.masterless') %}
```
giorgiosironi: makes sense as a condition

elife/masterless.sls
```diff
@@ -1,8 +1,64 @@
...
         - name: salt-minion
         - enable: false
-        - onlyif:
-            # running within vagrant
-            - test -d /vagrant
+
+# deny access to all except those in allow.all and allow.master-server
+# necessary to ensure those with access don't inadvertently grant access to others
+
+# zero out all allowed keys
```
giorgiosironi: I think you want this to run after the `elife/deploy-user.sls`, so it does need some `require` statement?

elife/masterless.sls
```diff
@@ -1,8 +1,64 @@
...
+deny_all_access:
+    cmd.run:
+        - name: |
+            echo > /home/elife/.ssh/authorized_keys
+            # bootstrap user is needed to login
+            # masterless environment disables adding any other keys to bootstrap user
+            #echo > /home/ubuntu/.ssh/authorized_keys
```
giorgiosironi: why is the second `echo` commented out?

giorgiosironi: figured out from https://github.com/elifesciences/builder-private/pull/34/files but I would remove the `echo` at all then

elife/masterless.sls
```diff
@@ -1,8 +1,64 @@
...
+{% for username in allowed %}
+    {% if pillar.elife.ssh_users.has_key(username) %}
+
+masterless-ssh-access-for-{{ username }}:
+    ssh_auth.present:
+        - user: {{ pillar.elife.deploy_user.username }}
+        - name: {{ pillar.elife.ssh_users[username] }}
+        - comment: {{ username }}
+        - require:
+            - cmd: /home/{{ user }}/.ssh/
```
giorgiosironi: also here `deny_all_access` could be `require`d

lsh-0:
dependencies are all hooked up now. merging.

PR #64, opened 2017-09-11
gearman, changes to default config

lsh-0:


elife/gearman.sls
```diff
@@ -15,6 +15,20 @@ gearman-upstart-script:
...
+gearman-systemd-script:
+    file.managed:
+        - name: /lib/systemd/system/gearman-job-server.service
+        - source: salt://elife/config/lib-systemd-system-gearman-job-server.service
+        - require:
+            - gearman-daemon
+
+default-gearman-config:
+    file.managed:
+        - name: /etc/gearman.conf
```
giorgiosironi: this state conflicts with https://github.com/elifesciences/search-formula/pull/20/files#diff-241a7d358f42c071e2e8fcb30afa3cf1R26 as they manage the same file. We tried doing that with elife-bot-formula and redis for the same reason but eventually failed to work and was replaced with another approach:
https://github.com/elifesciences/elife-bot-formula/blob/master/salt/pillar/elife-bot.sls#L5-L7
I suggest the same for Gearman, to only manage the file here and to switch its configuration to persistent or not depending on the pillar (which can also provide the credentials for the database when persistence is enabled).

elife/config/etc-init-gearman-job-server.conf
```diff
@@ -1,16 +1,8 @@
...
-#start on (filesystem and net-device-up IFACE=lo)
 start on postgresql-ready
 stop on runlevel [!2345]
 setuid gearman
-
 respawn
-
-script
-    . /etc/default/gearman-job-server
-    exec /usr/sbin/gearmand --log-file=/var/log/gearman-job-server/gearman.log $PARAMS
```
giorgiosironi: good to put the log file in the configuration

elife/gearman.sls
lsh-0: sure, we can do that.

lsh-0:
https://github.com/elifesciences/builder-private/pull/62

giorgiosironi:
builder-private configuration is in place. The conflict here is that `elife.gearman` has been split into client (just the PHP extension) and server, so the changes should be applied to the server part. This would unblock https://github.com/elifesciences/search-formula/pull/20 which needs to be merged, since we made lots of changes to search to load balance it and can only diverge more.

PR #67, opened 2017-09-26
Fixes from heavybox

giorgiosironi:
- Removed unused, not working `elasticsearch`
- Removed unused, not working `memcached`
- Removed deprecated New Relic state files that we don't use anymore
- Differentiate nodejs versions state names


lsh-0:
that `elasticsearch.sls` file is some old old code. part of the original stuff in the original builder.

PR #70, opened 2017-10-09
Bugfix gearman ppa

lsh-0:
this should fix it for both trusty and xenial

### DIFF
--- a/elife/gearman.sls
+++ b/elife/gearman.sls
@@ -1,3 +1,26 @@
+
+{% if 
+    salt['file.file_exists']('/etc/apt/sources.list.d/ondrej-ubuntu-php-trusty.list') or
+    salt['file.file_exists']('/etc/apt/sources.list.d/ondrej-ubuntu-php-xenial.list')
+%}
+
+# we're using a third party ppa for php and need the separate ppa for gearman
+
+php-ppa-gearman:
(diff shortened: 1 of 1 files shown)


lsh-0:
ah! I can see Alfred at work now. cool. I'll merge it if it passes

lsh-0:
eh. failed. testing a fresh install manually

lsh-0:
can't replicate Alfred's problem on a fresh install, my fix works.

@giorgiosironi could you please merge this at your convenience.

giorgiosironi:
I don't know how this works: I thought this kind of file checking is executed at YAML rendering time. If that was it, the files shouldn't be there as the other Salt state adding the `ondrej` repositories has not been executed yet.

But since it works experimentally, let's merge.

giorgiosironi:
It seems sometimes we want to run a build from scratch on a branch rather than building with the existing formula and then switch to the new commit. An idea for the future, if `builder` allows it.

lsh-0:
> I don't know how this works: I thought this kind of file checking is executed at YAML rendering time.

that's a good point. it was a check that started off as one thing, then became another, then two checks and suddenly you have something that should be done at runtime rather than compiletime. If it causes problems, we can add an elaborate `onlyif`

PR #71, opened 2017-10-10
gearman, shifted file check for ppa state into state requisite

lsh-0:


giorgiosironi:
Building a local search instance to check this (as we need a build from scratch).

giorgiosironi:
My update failed due to connection problems, but these states were successful.

PR #72, opened 2017-10-19
Java 8 temporary fix

giorgiosironi:
See linked posts, upon Java 9 coming out Oracle removed some installers breaking everyone automating the installation of Java 8. We should switch to OpenJDK in the future if possible


giorgiosironi:
Should be red because the current master is broken, and this is the fix.

giorgiosironi:
Works also on reinstalling (apt-get is idempotent and the `sed` commands do nothing).

PR #73, opened 2017-10-23
Some Postgres setup is not necessary on RDS

giorgiosironi:
- Do not run the local daemon (but still install all packages)
- Do not deploy init scripts
- Do not deploy server configuration


giorgiosironi:
The `heavybox` base failed on `goaws-install`.

giorgiosironi:
...which does fail also locally. These PRs are becoming useful for the wrong goal: finding out breakages in existing formulas.

giorgiosironi:
So the commit of `goaws` was pinned, but its dependencies weren't pinned (`composer.lock`/`requirements.txt` anyone?). We need to upgrade our Go version 7 minor releases ahead, because some random library that cannot be pinned decided to rely on a newer language feature, which is totally unacceptable. I'm more inclined to try using a fixed version of the Docker container and banning Go from being used at all.

giorgiosironi:
Now green after yak shaving Go applications.

lsh-0:
oh, sorry - I overlooked the 'not' in those clauses. this will still break scheduler as it uses a local psql instance while the dashboard uses rds.

giorgiosironi:
The underlying reason is not to mislead people encountering a local Postgres when RDS is used instead, while also the RAM used by the server can be freed to be used by the applications.

`elife-article-scheduler` should be the only blocker then:
<img src="https://vignette.wikia.nocookie.net/shrek/images/1/16/Grrr._You_made_the_cat_angry.jpg/revision/latest?cb=20120215093600" alt="elife-article-scheduler, you have made the cat angry"

lsh-0:
nice ;) If it's not already, I'll open a PR on the dashboard formula to ensure local psql is running. Then we can merge this afterwards.

lsh-0:
https://github.com/elifesciences/elife-dashboard-formula/pull/2

giorgiosironi:
Still propagating some changes to the dashboard, but should be merged on Monday.

PR #74, opened 2017-10-27
ELPP-3237 wait_packagist cleans selectively Composer's cache

giorgiosironi:
The script doesn't use a cache by itself, but it cleans the Composer's relevant cache file so that if subsequent `composer` commands are issued they will see the same version of the package rather then an old one.

### DIFF
--- a/elife/jenkins-scripts/wait_packagist.sh
+++ b/elife/jenkins-scripts/wait_packagist.sh
@@ -22,3 +22,7 @@ while : ; do
         break
     fi
 done
+
+cache_file_for_package_revisions=~/.cache/composer/repo/https---packagist.org/provider-"${package/\//\$}".json
+echo "Removing $cache_file_for_package_revisions to make sure Composer will see the update."
+rm -f "$cache_file_for_package_revisions"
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Which repository has decided to break today?
`Failed to fetch http://ppa.launchpad.net/ondrej/php/ubuntu/dists/trusty/main/binary-amd64/Packages`

elife/jenkins-scripts/wait_packagist.sh
```diff
@@ -22,3 +22,7 @@ while : ; do
         break
     fi
 done
+
+cache_file_for_package_revisions=~/.cache/composer/repo/https---packagist.org/provider-"${package/\//\$}".json
```
thewilkybarkid: `/usr/lib/composer/.cache` (https://github.com/elifesciences/builder-base-formula/blob/master/elife/composer.sls#L5 - might want to move to pillar?)

giorgiosironi: I'd like to keep this running as-is rather than making it a template for a script, so I'll try an environment variable

PR #76, opened 2017-10-31
ELPP-3235 uwsgi and systemd template

giorgiosironi:


giorgiosironi:
Ah, failed on 14.04. Will introduce a guard clause.

giorgiosironi:
I thought if emperor mode gets introduced, it will still need a single uwsgi service to be configured, so it will benefit from the systemd template (it could get more general in the future).

lsh-0:
My plans for the emperor mode probably would have been configured similarly to how it's been done here. I think this is good enough for now.

PR #78, opened 2017-11-08
donotmerge - fixes bug where postgres.user_exists and postgres.db_exists don't exist

lsh-0:
happened on a new vagrant build of the elife-metrics project.

unless this PR is run on a machine built from scratch it's not going to tell us anything.


lsh-0:
sounds similar to this bug: https://github.com/saltstack/salt/issues/8485

lsh-0:
just double checked the logic. it may actually be a bug in salt, but it shouldn't affect anything as it only deals with a legacy db that needs it's permissions altered.

lsh-0:
I think this will affect brand new machines. Machines where postgresql is already installed should be fine

PR #79, opened 2017-11-08
Do not download docker-compose repeatedly

giorgiosironi:


thewilkybarkid:
Would using a `file.managed` be better for fetching a remote file? Like https://github.com/elifesciences/builder-base-formula/blob/58ed61488085d3eaa8d0bd41db75b15b2422ff25/elife/selenium.sls#L63-L67

(Upgrading looks hard without?)

giorgiosironi:
I'm not sure, because it would have to be calculating the md5 of the remote file and the only way to do that is by downloading it?
We may also use `wget -c` instead to cache the file locally.
Haven't upgrade it yet, but in that case the URL changes and the `unless` would have to change too to check the version.

thewilkybarkid:
I assumed it compared the md5 of the local file (if it exists) to know whether to download it again.

giorgiosironi:
Ah right, I confused the download of a .deb package with the download of a standalone binary, this is the latter so it *is* kept around after installation.

elife/docker.sls
```diff
@@ -23,11 +23,15 @@ docker-packages:
...
 
 docker-compose:
-    cmd.run:
-        - name: |
-            curl -L https://github.com/docker/compose/releases/download/1.16.1/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
-            chmod +x /usr/local/bin/docker-compose 
+    file.managed:
+        - name: /usr/local/bin/docker-compose 
+        - source: https://github.com/docker/compose/releases/download/1.16.1/docker-compose-Linux-x86_64
+        - source_hash: 0ed7666983387be3fd73d700c6627cdf
```
thewilkybarkid: `source_hash: md5=0ed7666983387be3fd73d700c6627cdf`

giorgiosironi: makes sense they don't assume a format

giorgiosironi: and the build did fail :tea:

giorgiosironi:
Finally working.

PR #80, opened 2017-11-09
fixes a problem I introduced with the rds-perms script

lsh-0:
this has been tested with restoring a backup afterwards and running tests without any permissions problems


giorgiosironi:
RDS permissions and creations always give me a headache, can we capture some of this information as an `adr/` file?
- is the database always supposed to be owned by the rds superuser or by the application user
- why is the rds superuser not a real superuser (because RDS does not give you full access, is there a documentation link)
- what should be done when a database needs to be reset? Can't really drop it

lsh-0:
I completely agree and empathize. This postgresql+rds permission thing has been torture. I'll put some words togther.

PR #83, opened 2017-11-10
Fix Composer cache path

thewilkybarkid:
https://alfred.elifesciences.org/job/dependencies-error-pages-update-patterns-php/261/ picked up the wrong commit, tried a few projects and the path seems to have been incorrect everywhere.


giorgiosironi:
May depend on the `composer` version? Is this consistent with what you had locally?

thewilkybarkid:
Matches locally too.

PR #84, opened 2017-11-13
ELPP-2550 Improve PHP config

thewilkybarkid:
We've not really touched much of the PHP configuration, there look to be areas we could improve.

### DIFF
--- a/elife/config/etc-php-7.0-fpm-php.ini
+++ b/elife/config/etc-php-7.0-fpm-php.ini
@@ -386,7 +386,7 @@ max_input_time = 60
 
 ; Maximum amount of memory a script may consume (128MB)
 ; http://php.net/memory-limit
-memory_limit = 128M
+memory_limit = {{ pillar.elife.php.memory_limit }}
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Error handling and logging ;
--- a/elife/config/etc-php-7.0-fpm-php.ini
+++ b/elife/config/etc-php-7.0-fpm-php.ini
@@ -329,7 +329,7 @@ disable_classes =
 ; file or directory. For systems with rarely changing files, consider increasing this
 ; value.
 ; http://php.net/realpath-cache-ttl
-;realpath_cache_ttl = 120
+realpath_cache_ttl = 3600
 
 ; Enables or disables the circular reference collector.
 ; http://php.net/zend.enable-gc
@@ -825,7 +825,7 @@ allow_url_include = Off
 
 ; Default timeout for socket based streams (seconds)
 ; http://php.net/default-socket-timeout
-default_socket_timeout = 60
+default_socket_timeout = 5
 
 ; If your scripts have to deal with files from Macintosh systems,
 ; or you are running on a Mac and need to deal with files from
@@ -1737,14 +1737,14 @@ ldap.max_links = -1
 ;opcache.enable_cli=0
 
 ; The OPcache shared memory storage size.
-;opcache.memory_consumption=64
+opcache.memory_consumption=512
 
 ; The amount of memory for interned strings in Mbytes.
-;opcache.interned_strings_buffer=4
+opcache.interned_strings_buffer=64
 
 ; The maximum number of keys (scripts) in the OPcache hash table.
 ; Only numbers between 200 and 100000 are allowed.
-;opcache.max_accelerated_files=2000
+opcache.max_accelerated_files=20000
 
 ; The maximum percentage of "wasted" memory until a restart is scheduled.
 ;opcache.max_wasted_percentage=5
@@ -1757,7 +1757,7 @@ ldap.max_links = -1
 
 ; When disabled, you must reset the OPcache manually or restart the
 ; webserver for changes to the filesystem to take effect.
-;opcache.validate_timestamps=1
+opcache.validate_timestamps={% if pillar.elife.dev %}1{% else %}0{% endif %}
 
 ; How often (in seconds) to check file timestamps for changes to the shared
 ; memory storage allocation. ("1" means validate once per second, but only
--- a/elife/config/etc-php-7.0-fpm-php.ini
+++ b/elife/config/etc-php-7.0-fpm-php.ini
@@ -1757,7 +1757,7 @@ opcache.max_accelerated_files=20000
 
 ; When disabled, you must reset the OPcache manually or restart the
 ; webserver for changes to the filesystem to take effect.
-opcache.validate_timestamps={% if pillar.elife.dev %}1{% else %}0{% endif %}
+;opcache.validate_timestamps=1
 
 ; How often (in seconds) to check file timestamps for changes to the shared
 ; memory storage allocation. ("1" means validate once per second, but only
--- a/elife/config/etc-php-7.0-fpm-pool.d-www.conf
+++ b/elife/config/etc-php-7.0-fpm-pool.d-www.conf
@@ -95,6 +95,22 @@ listen.group = www-data
 ; Note: This value is mandatory.
 pm = dynamic
 
+{% if pillar.elife.env in ['prod', 'end2end'] %}
+    {% set available_memory = salt['grains.get']('mem_total') * 0.8 %}
+{% else %}
+    {% set available_memory = salt['grains.get']('mem_total') * 0.25 %}
+{% endif %}
+{% set max_children = (available_memory / pillar.elife.php.memory_limit[:-1]|float)|round(0, 'floor')|int %}
+{% if max_children < 5 %}{% set max_children = 5 %}{% endif %}
+
+{% set min_spare_servers = (max_children * 0.1)|round|int %}
+{% if min_spare_servers < 1 %}{% set min_spare_servers = 1 %}{% endif %}
+
+{% set max_spare_servers = (max_children * 0.3)|round|int %}
+{% if min_spare_servers < 3 %}{% set max_spare_servers = 3 %}{% endif %}
+
+{% set start_servers = (min_spare_servers + (max_spare_servers - min_spare_servers) / 2)|round|int %}
+
 ; The number of child processes to be created when pm is set to 'static' and the
 ; maximum number of child processes when pm is set to 'dynamic' or 'ondemand'.
 ; This value sets the limit on the number of simultaneous requests that will be
@@ -104,22 +120,22 @@ pm = dynamic
 ; forget to tweak pm.* to fit your needs.
 ; Note: Used when pm is set to 'static', 'dynamic' or 'ondemand'
 ; Note: This value is mandatory.
-pm.max_children = 30
+pm.max_children = {{ max_children }}
 
 ; The number of child processes created on startup.
 ; Note: Used only when pm is set to 'dynamic'
 ; Default Value: min_spare_servers + (max_spare_servers - min_spare_servers) / 2
-pm.start_servers = 5
+pm.start_servers = {{ start_servers }}
 
 ; The desired minimum number of idle server processes.
 ; Note: Used only when pm is set to 'dynamic'
 ; Note: Mandatory when pm is set to 'dynamic'
-pm.min_spare_servers = 3
+pm.min_spare_servers = {{ min_spare_servers }}
 
 ; The desired maximum number of idle server processes.
 ; Note: Used only when pm is set to 'dynamic'
 ; Note: Mandatory when pm is set to 'dynamic'
-pm.max_spare_servers = 5
+pm.max_spare_servers = {{ max_spare_servers }}
 
 ; The number of seconds after which an idle process will be killed.
 ; Note: Used only when pm is set to 'ondemand'
--- a/elife/config/etc-php-7.0-fpm-pool.d-www.conf
+++ b/elife/config/etc-php-7.0-fpm-pool.d-www.conf
@@ -95,21 +95,21 @@ listen.group = www-data
 ; Note: This value is mandatory.
 pm = dynamic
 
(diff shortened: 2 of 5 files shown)


elife/config/etc-php-7.0-fpm-php.ini
```diff
@@ -825,7 +825,7 @@ allow_url_include = Off
 
 ; Default timeout for socket based streams (seconds)
 ; http://php.net/default-socket-timeout
-default_socket_timeout = 60
+default_socket_timeout = 5
```
thewilkybarkid: Long timeouts should be set explicitly.

elife/config/etc-php-7.0-fpm-php.ini
```diff
@@ -1757,7 +1757,7 @@ ldap.max_links = -1
 
 ; When disabled, you must reset the OPcache manually or restart the
 ; webserver for changes to the filesystem to take effect.
-;opcache.validate_timestamps=1
+opcache.validate_timestamps={% if pillar.elife.dev %}1{% else %}0{% endif %}
```
thewilkybarkid: This means that any PHP application will have to restart PHP-FPM for changes to be recognised.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+; The number of child processes to be created when pm is set to 'static' and the
+; maximum number of child processes when pm is set to 'dynamic' or 'ondemand'.
+; This value sets the limit on the number of simultaneous requests that will be
+; served. Equivalent to the ApacheMaxClients directive with mpm_prefork.
+; Equivalent to the PHP_FCGI_CHILDREN environment variable in the original PHP
+; CGI. The below defaults are based on a server without much resources. Don't
+; forget to tweak pm.* to fit your needs.
+; Note: Used when pm is set to 'static', 'dynamic' or 'ondemand'
+; Note: This value is mandatory.
+pm.max_children = 30
```
thewilkybarkid: Originally the default of 5.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+
+; The number of child processes created on startup.
+; Note: Used only when pm is set to 'dynamic'
+; Default Value: min_spare_servers + (max_spare_servers - min_spare_servers) / 2
+pm.start_servers = 5
```
thewilkybarkid: Originally the default of 2.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+
+; The desired minimum number of idle server processes.
+; Note: Used only when pm is set to 'dynamic'
+; Note: Mandatory when pm is set to 'dynamic'
+pm.min_spare_servers = 3
```
thewilkybarkid: Originally the default of 1.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+
+; The desired maximum number of idle server processes.
+; Note: Used only when pm is set to 'dynamic'
+; Note: Mandatory when pm is set to 'dynamic'
+pm.max_spare_servers = 5
```
thewilkybarkid: Originally the default of 3.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+; The number of seconds after which an idle process will be killed.
+; Note: Used only when pm is set to 'ondemand'
+; Default Value: 10s
+;pm.process_idle_timeout = 10s;
+
+; The number of requests each child process should execute before respawning.
+; This can be useful to work around memory leaks in 3rd party libraries. For
+; endless request processing specify '0'. Equivalent to PHP_FCGI_MAX_REQUESTS.
+; Default Value: 0
+pm.max_requests = 1000
```
thewilkybarkid: Originally the default of 500.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+;                                    of 'idle' processes is greater than this
+;                                    number then some children will be killed.
+;  ondemand - no children are created at startup. Children will be forked when
+;             new requests will connect. The following parameter are used:
+;             pm.max_children           - the maximum number of children that
+;                                         can be alive at the same time.
+;             pm.process_idle_timeout   - The number of seconds after which
+;                                         an idle process will be killed.
+; Note: This value is mandatory.
+pm = dynamic
```
thewilkybarkid: `static` looks to be the optimal value, but needs careful analysis.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
thewilkybarkid: This is based on a rough calculation of Journal on its `t2.small` (then rounded down a bit). Might not be too high for some (Drupal?).

thewilkybarkid: Maybe parametrise? (Or calculate?)

elife/config/etc-php-7.0-fpm-php.ini
giorgiosironi: Do we have any idea of where this is used? I think curl, being its own extensions, may not be influenced by this?

thewilkybarkid: Not cURL, but `file_get_contents()` etc.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
```diff
@@ -0,0 +1,415 @@
...
+; This directive may be used to customize the response of a ping request. The
+; response is formatted as text/plain with a 200 response code.
+; Default Value: pong
+;ping.response = pong
+
+; The access log file
+; Default: not set
+;access.log = log/$pool.access.log
+
+{% raw -%}
```
thewilkybarkid: Jinja not actually used yet, but to avoid the problem happening in the future...

elife/config/etc-php-7.0-fpm-php.ini
giorgiosironi: We shouldn't modify files manually very often.
During deploy/Salt highstates, https://github.com/elifesciences/builder-base-formula/blob/master/elife/nginx-php7.sls does not restart the service by default. States that update the files need a `watch_in` to trigger the restart https://github.com/elifesciences/journal-formula/blob/master/salt/journal/init.sls#L14-L15 before this can change can be made.

elife/nginx-php7.sls
```diff
@@ -34,4 +42,5 @@ php-fpm:
         - name: php7.0-fpm
         - require:
             - file: php-fpm
-            - file: php-fpm-config            
+            - file: php-fpm-config
+            - php-fpm-pool
```
giorgiosironi: probably needs 
```
- watch:
    - php-fpm-pool
```
to be restarted on changes. Probably should also watch the other ones, same reasoning.
Alternative could be to try `reload: True`.

giorgiosironi: Example: https://github.com/Enrise/phpfpm-formula/blob/1b989c5913b3f437100982c73227aaeeb6ed18d0/phpfpm/service.sls#L27-L33

giorgiosironi: Also `enable: True` would make sure it's started on boot.

thewilkybarkid: Added.

thewilkybarkid:
Build failures look unrelated.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
giorgiosironi: In general this is usually proportional to the number of CPU (cores), so it should be a multiple of ``{% salt['grains.get']('num_cpus') %}``, which is `1` on `t2.small`. Same for other related values like `pm.start_servers` maybe.

elife/nginx-php7.sls
```diff
@@ -32,6 +40,11 @@ php-fpm:
 
     service.running:
         - name: php7.0-fpm
+        - enable: True
         - require:
             - file: php-fpm
-            - file: php-fpm-config            
+            - file: php-fpm-config
+            - php-fpm-pool
+        - watch:
```
giorgiosironi: also `pkg: php-nginx-deps` if the package version changes for some reason?

giorgiosironi:
Trying manually to check what's happening.

giorgiosironi:
We had a race condition between the two commits, who tried to get build using an instance with the same name. Going to add a lock on that.

elife/nginx-php7.sls
```diff
@@ -32,6 +40,27 @@ php-fpm:
...
+            - file: php-fpm-config
+            - file: php-fpm-pool
+
+php-cachetool:
+    file.managed:
+        - name: /usr/local/bin/cachetool
+        - source: https://gordalina.github.io/cachetool/downloads/cachetool.phar
+        - source_hash: md5=fa7ce33b37dba2642329b9a6bdc720b1
```
thewilkybarkid: Looks like this will break on every release, since there's no version in the URI...

elife/config/etc-php-7.0-fpm-pool.d-www.conf
thewilkybarkid: Advice seems to be available memory ÷ memory required for a request.

giorgiosironi: Memory is definitely an hard limit, `{% salt['grains.get']('mem_total') %}` is `2000` on `journal` servers.

elife/nginx-php7.sls
giorgiosironi: It defines a `bin` folder so a `composer global require ...` could be more helpful in managing its version? https://github.com/gordalina/cachetool/blob/master/composer.json#L35
It would probably be slower than a phar and an MD5 check.

giorgiosironi: Or https://github.com/gordalina/cachetool/releases to get a tarball.

thewilkybarkid: Don't hugely want to get involved in a more complex build process for a little tool. Could open an issue upstream and place the PHAR on S3 for now?

thewilkybarkid: Global Composer dependencies aren't great, due to constraint conflicts.

thewilkybarkid: I suggest we have this deployed first, and use it to try and optimise values (by looking at what's actually happening).

giorgiosironi: :+1: for as-is and .phar on S3 for the medium-term, may not even prove useful.

thewilkybarkid: Moved to #85.

thewilkybarkid:
I've been through the PHP servers, and as expected Journal/Journal CMS hit `max_accelerated_files`. Nothing hits `memory_consumption`, but that's related (we may not need it to go as high as suggested, but we can tweak later).

Not sure we actually need to change `interned_strings_buffer`, Journal CMS is closest to hitting the current max (97% full). Maybe just double it.

elife/config/etc-php-7.0-fpm-pool.d-www.conf
thewilkybarkid: `memory_limit` is currently `128M`, so combined that's more than available for Journal. Though I don't believe requests get anywhere near that limit, so could be reduced. Looking to see if can find out how much memory requests use.

thewilkybarkid: Think we'd have to add a custom parameter to New Relic then learn how to use Insights.

giorgiosironi: Didn't find a way to do that in New Relic Infrastructure, the metrics being reported are not configurable.

thewilkybarkid: Maybe just set `memory_limit` in Pillar, then calculate this as `(mem_total × .8) ÷ memory_limit` (on basis that other services are run elsewhere). Would give 12 for Journal, which is definitely better (`memory_limit` _is_ far too high there, locally it looks like the archive pages use the most memory (not paginated), <40).

thewilkybarkid: Article 06416 uses 48MB locally (giant list of authors), could well be the biggest article JSON?

giorgiosironi: `memory_limit` can be tuned for journal but we need a large default to prevent, for example, journal-cms from crashing on half the API calls. The pillar should go into `pillar.elife` so it's easy to override.

thewilkybarkid: I'd go the other way, make it 64MB by default then increase where needed.

giorgiosironi: As long as the risky projects have an override from the get-go, `journal-cms` for sure, the others should be fine.

giorgiosironi: To be added here: https://github.com/elifesciences/builder-private/blob/master/pillar/journal-cms.sls#L49

elife/nginx-php7.sls
thewilkybarkid: Added.

PR #87, opened 2017-11-21
Support PHP long-running processes in 16.04

giorgiosironi:
Purely through pillar configuration, since we are already tired of creating systemd service files.

I think this can be extended to Python (or generic) processes later, but with customization of the template to allow for things like New Relic coreography or custom options, so not trivial.

### DIFF
--- a/elife/php-long-running-processes.sls
+++ b/elife/php-long-running-processes.sls
@@ -15,7 +15,9 @@ php-long-running-process-service-{{ hyphenized }}:
             command: {{ configuration.command }}
         - require:
             - php
-            # TODO: add optional require
+            {% if configuration.get('require', None) -%}
+            - {{ configuration.get('require') }}
+            {%- endif %}
         - require_in:
             - cmd: php-long-running-processes-load-configuration
 {% endfor %}
--- a/elife/php-long-running-processes.sls
+++ b/elife/php-long-running-processes.sls
@@ -30,7 +30,7 @@ php-long-running-processes-load-configuration:
 {% set hyphenized = name | replace('_', '-') %}
 {% set service_name = salt['elife.project_name']() + '-' + hyphenized %}
 {% for i in range(1, configuration['number'] + 1) %}
-php-long-running-process-service-{{ hyphenized }}-{{ i }}-start:
+php-long-running-process-service-{{ hyphenized }}-{{ i }}-enable:
     cmd.run:
         - name: systemctl enable {{ service_name }}@{{ i }}
         - require:
@@ -44,6 +44,4 @@ php-long-running-process-service-{{ hyphenized }}-parallel-restart:
     cmd.run:
         - name: systemctl restart {{ service_name }}@{1..{{ configuration['number'] }}}
 {% endfor %}
-
-    
 {% endif %}
--- a/elife/php-long-running-processes.sls
+++ b/elife/php-long-running-processes.sls
@@ -27,9 +27,8 @@ php-long-running-processes-load-configuration:
 {% for name, configuration in pillar.elife.php.processes.configuration.items() %}
 {% set hyphenized = name | replace("_", "-") %}
 {% set service_name = salt['elife.project_name']() + "-" + hyphenized %}
-# TODO: add counter and instance numbers
 # TODO: " vs '
-{% for i in range(0, configuration['number']) %}
+{% for i in range(1, configuration['number'] + 1) %}
 php-long-running-process-service-{{ hyphenized }}-{{ i }}-start:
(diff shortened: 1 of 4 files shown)


elife/php-long-running-processes.sls
```diff
@@ -0,0 +1,47 @@
...
+        - source: salt://elife/config/lib-systemd-system-php-service.service
+        - template: jinja
+        - context:
+            name: {{ hyphenized }}
+            folder: {{ configuration.folder }}
+            command: {{ configuration.command }}
+        - require:
+            - php
+            {% if configuration.get('require', None) -%}
+            - {{ configuration.get('require') }}
```
thewilkybarkid: Does this allow for only 1 requirement?

elife/php-long-running-processes.sls
```diff
@@ -0,0 +1,47 @@
...
+php-long-running-process-service-{{ hyphenized }}-{{ i }}-enable:
+    cmd.run:
+        - name: systemctl enable {{ service_name }}@{{ i }}
+        - require:
+            - php-long-running-process-service-{{ hyphenized }}
+        - require_in:
+            - file: php-long-running-process-service-{{ hyphenized }}-parallel-restart
+{% endfor %}
+
+# TODO: this could be a single command for all kinds of processes
```
thewilkybarkid: 👍

elife/php-long-running-processes.sls
```diff
@@ -0,0 +1,47 @@
+# only supports 16.04!
+
+{% if pillar.elife.php.processes.enabled %}
+{% for name, configuration in pillar.elife.php.processes.configuration.items() %}
+{% set hyphenized = name | replace('_', '-') %}
+{% set service_name = salt['elife.project_name']() + '-' + hyphenized %}
+php-long-running-process-service-{{ hyphenized }}:
```
thewilkybarkid: That's quite a mouthful...

thewilkybarkid: `php-long-running-process` looks redundant?

giorgiosironi: Want to absolutely avoid these states clashing with the ones of a project import this, any shorter unique prefix to use?

elife/php-long-running-processes.sls
giorgiosironi: Yes, usually this is composer install or the configuration file provisioning, but integrates the two files on a single state. A project can still create a `cmd.run` state to converge its disparate states into one, rather than propagating multiple state names to builder-private or its pillar.

elife/php-long-running-processes.sls
giorgiosironi: I think they can all put in the same `systemctl restart` call. But YAGNI at the moment, as multiple types of processes are still supported but restarts not as fast as possible :thinking:

elife/php-long-running-processes.sls
thewilkybarkid: Using https://github.com/elifesciences/builder-base-formula/pull/87/files#diff-ab2b69d66c8c8bcba9fd1952e9ab976aR177, `{{ hyphenized }}` is just `annotations` right? That doesn't look like the best name, should `annotations-queue-watch`? Prefixing that with `service` should be ok?

elife/php-long-running-processes.sls
```diff
@@ -0,0 +1,47 @@
...
+    file.managed:
+        - name: /lib/systemd/system/{{ service_name }}@.service
+        - source: salt://elife/config/lib-systemd-system-php-service.service
+        - template: jinja
+        - context:
+            name: {{ hyphenized }}
+            folder: {{ configuration.folder }}
+            command: {{ configuration.command }}
+        - require:
+            - php
```
thewilkybarkid: Is this the only thing here that makes it PHP-specific?

elife/php-long-running-processes.sls
giorgiosironi: Yeah, that should be `queue_watch` (the fact that pillar keys require `_` is the reason for the hyphenization)

elife/php-long-running-processes.sls
giorgiosironi: Also we won't get away with https://github.com/elifesciences/builder-base-formula/pull/87/files#diff-568b5141ae75a6b187f83bb22ee62d04R14 for Python as it requires New Relic magic around it:
https://github.com/elifesciences/profiles-formula/blob/master/salt/profiles/config/lib-systemd-system-uwsgi-profiles.service#L15
possibly `KillSignal` and `Type` need configurability too

PR #88, opened 2017-11-22
Supporting generic *-dummy services for testing

giorgiosironi:
Extracted from `profiles.orcid-dummy`.

### DIFF
--- a/elife/php-dummies.sls
+++ b/elife/php-dummies.sls
@@ -44,7 +44,7 @@
 {{ name }}-composer-install:
     cmd.run:
         {% if pillar.elife.env != 'dev' %}
-        - name: composer --no-interaction install --no-suggest --classmap-authoritative
+        - name: composer --no-interaction install --no-suggest --no-dev --classmap-authoritative
         {% else %}
         - name: composer --no-interaction install --no-suggest
         {% endif %}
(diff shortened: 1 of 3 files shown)


elife/php-dummies.sls
```diff
@@ -0,0 +1,72 @@
...
+        - name: git checkout $(cat {{ configuration['pinned_revision_file'] }})
+        - cwd: /srv/{{ name }}
+        - user: {{ pillar.elife.deploy_user.username }}
+        - require:
+            - file: {{ name }}-repository
+
+{{ name }}-composer-install:
+    cmd.run:
+        {% if pillar.elife.env != 'dev' %}
+        - name: composer --no-interaction install --no-suggest --classmap-authoritative
```
thewilkybarkid: `--no-dev`?

elife/php-dummies.sls
```diff
@@ -0,0 +1,72 @@
...
+        {% else %}
+        - name: composer --no-interaction install --no-suggest
```
thewilkybarkid: Probably ok to separate `dev`, you might want to tweak the dummy at the same time as the main project.

giorgiosironi: Least astonishment, ok

giorgiosironi:
Well, composer and the nginx setup are PHP-specific so trying to use it with other dummies will break.

PR #89, opened 2017-11-24
Small Docker support improvements

giorgiosironi:


elife/docker.sls
```diff
@@ -35,3 +35,10 @@ docker-compose:
         - require:
             - file: docker-compose
 
+docker-deploy-user-in-group:
```
giorgiosironi: avoids the use of `sudo`

PR #91, opened 2017-12-01
Pin Composer version temporarily

thewilkybarkid:
Composer 1.5.3 seems to have a problem executing https://github.com/elifesciences/builder-base-formula/blob/688ff004ce22ce3e1b8d0ca53d6ee9f59484bedc/elife/proofreader-php.sls#L1-L6

I've opened https://github.com/composer/composer/issues/6860, but temporarily we can pin the version to the previous release which works.

### DIFF
--- a/elife/composer.sls
+++ b/elife/composer.sls
@@ -79,11 +79,14 @@ composer-global-paths:
 
 update-composer:
     cmd.run:
-        - name: composer self-update
-        - onlyif:
-            - which composer
-        - require_in:
+        - name: composer self-update 1.5.2
+        - require:
             - cmd: install-composer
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
There's no `pr-update` run @giorgiosironi, is that because `pr-base` failed?

thewilkybarkid:
I've tested locally (both updating and with a fresh VM) and it works.

giorgiosironi:
Yes, `pr-base` is the highstate with the master of the formula, necessary to apply the change `pr-update` (`pr-head` is the whole build). Pending some builder improvements we will add `pr-fresh` which would create in parallel a VM using the new formula directly.

PR #92, opened 2017-12-01
Revert "Pin Composer version temporarily"

thewilkybarkid:
Reverts elifesciences/builder-base-formula#91

Composer 1.5.5 fixes the problem (1.5.4 was released, but didn't work).


giorgiosironi:
Should we pin the new version instead?

thewilkybarkid:
Probably not, unless you fancy maintaining it. (First time it's had a problem.)

PR #95, opened 2017-12-07
[wip] installs and configures pgbouncer.

lsh-0:
* configuration is at pillar.elife.postgresql.pool

With pgbouncer as the connection pooling proxy, all apps using the postgresql-appdb state could be switched over to use it by having `pillar.elife.postgresql.host` and `port` point to localhost (pgbouncer)

connection pooling will reduce the connection overhead in all django applications. flask/other apps using postgresql may also benefit.

pinging @giorgiosironi for thoughts.

### DIFF
--- a/elife/config/etc-pgbouncer-pgbouncer.ini
+++ b/elife/config/etc-pgbouncer-pgbouncer.ini
@@ -1,0 +1,13 @@
+# https://pgbouncer.github.io/config.html
+
+[databases]
+{{ db_name }} = host={{ host }} dbname={{ db_name }} user={{ app_user }} password={{ app_pass }}
+
+[pgbouncer]
+pool_mode = session
+listen_addr = {{ pillar.elife.postgresql.pool.host }}
+listen_port = {{ pillar.elife.postgresql.pool.port }}
+auth_type = md5
+auth_file = users.txt
+logfile = /var/log/postgresql/pgbouncer.log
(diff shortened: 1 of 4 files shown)


giorgiosironi:
The concept is sound, and the trend is to [get some of these concerns out of the applications](https://cloudplatform.googleblog.com/2017/05/istio-modern-approach-to-developing-and.html) as they become smaller.

I don't know the particular technology but it's a backward compatible change so it can be rolled out on one service/environment at a time.

elife/config/etc-pgbouncer-pgbouncer.ini
```diff
@@ -0,0 +1,13 @@
+# https://pgbouncer.github.io/config.html
+
+[databases]
+{{ db_name }} = host={{ host }} dbname={{ db_name }} user={{ app_user }} password={{ app_pass }}
+
+[pgbouncer]
```
giorgiosironi: is there a maximum number of connections to configure? We haven't had any problem due to overprovisioning resources though.

lsh-0: there are a heap of options for fine tuning: https://pgbouncer.github.io/config.html

I think we'll be fine with the defaults, but I haven't actually tested this yet. It's just one of those things that has been nagging me for a while - that new-connection-per-query that Django does - and it seems like an easy win when combined with the psql-appdb work.

PR #96, opened 2017-12-11
improved ssl config for apache2

lsh-0:


elife/config/etc-apache2-conf-enabled-ssl.conf
```diff
@@ -0,0 +1,34 @@
...
+# note! these are quite strict rules that will exclude older browsers:
+#SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH
+# and these are more backwards compatible:
+SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:ECDHE-RSA-AES128-SHA:DHE-RSA-AES128-GCM-SHA256:AES256+EDH:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-SHA256:AES128-SHA256:AES256-SHA:AES128-SHA:DES-CBC3-SHA:HIGH:!aNULL:!eNULL:!EXPORT:!DES:!MD5:!PSK:!RC4
+
+SSLProtocol All -SSLv2 -SSLv3 -TLSv1 -TLSv1.1
+SSLHonorCipherOrder On
+
+Header always set Strict-Transport-Security "max-age=63072000; includeSubDomains; preload"
+Header always set X-Frame-Options DENY
```
giorgiosironi: doesn't use frames, I think (`crm` is the only project this should touch)

lsh-0:
sorry, I should have linked to the ticket: https://elifesciences.atlassian.net/browse/ELPP-2422

it has reports for both before and after

elife/config/etc-apache2-conf-enabled-ssl.conf
lsh-0: crm is the only project using apache.

any project actually using frames in their html probably should not. if they do, we can relax it from `DENY` to `SAMEORIGIN`

PR #98, opened 2017-12-13
ELPP-3213 Disable security update cron when managed externally

giorgiosironi:
In these environments, we are running https://alfred.elifesciences.org/blue/organizations/jenkins/process-daily-updates/activity and the two clash

### DIFF
--- a/elife/daily-system-updates.sls
+++ b/elife/daily-system-updates.sls
@@ -18,7 +18,6 @@ daily-system-update-log-rotater:
         - name: /etc/logrotate.d/daily-system-update
         - source: salt://elife/config/etc-logrotate.d-daily-system-update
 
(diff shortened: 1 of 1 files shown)


giorgiosironi:
This was not a problem until https://github.com/elifesciences/builder-base-formula/pull/97 reduced the random sleep time to a low value. Now it starts along with the bootstrap of the EC2 instance. Needs to eventually be propagated to all environments.

giorgiosironi:
After this, I think we need to recreate the AMIs for both 14.04 and 16.04 as new instances created from them will start the update before we can change the configuration with an highstate.

giorgiosironi:
Need to go with this to fix (if it works) formula builds.

PR #99, opened 2017-12-15
replaced a (very old) postfix prod instance check with preferred method

lsh-0:
note: is_prod_instance checked if env was either 'master' or 'prod', but postfix has never been used with the master server
note: 'ci' requires the service to be running and I'm not sure why. are we sending test emails somewhere/how? this might be better handled in pillar data

### DIFF
--- a/elife/postfix.sls
+++ b/elife/postfix.sls
@@ -5,9 +5,8 @@ postfix-mailserver:
             - mailutils # gives us the 'mail' command
 
     # mailserver only runs in production env to prevent accidental mailouts
-    {% if salt['elife.cfg']('project.is_prod_instance') %}
-    service.running:
-    {% elif pillar.elife.env == 'ci' %}
+    # why is mailserver running in ci ... ?
+    {% if pillar.elife.env in ['prod', 'ci'] %}
     service.running:
     {% else %}
     service.dead:


giorgiosironi:
Stopping ci emails is only setup for Civi: https://github.com/elifesciences/builder-private/blob/master/pillar/crm-ci.sls
So I agree on the removal of `is_prod_instance` but not sure on allowing `ci` since e.g. `journal-cms--ci` or whatever other service could send emails? Have to check where postfix is used.

lsh-0:
I've taken a look at `builder-private/salt/top.sls` and it looks like only `crm` and `alfred` are using postfix. The Alfred instance has a `--prod` suffix but I could have sworn it lived in `ci`. It doesn't matter because alfred [has an override](https://github.com/elifesciences/elife-alfred-formula/blob/master/salt/elife-alfred/postfix.sls)

I think it should be safe to remove `ci` from that list

giorgiosironi:
It's the `crm--ci` SMTP override, as you found out only `crm` is hitting this case:
https://github.com/elifesciences/builder-private/blob/master/pillar/crm-ci.sls
`elife-alfred` has been ported to the `prod` environment around the start of November.

giorgiosironi:
The context of `crm--ci` is people testing eTocs outside of the production server, and wanting to see the generated emails somewhere to check their design, links, etc.
If we expand the `crm--ci` pillar to the whole of `ci` we could then leave `ci` enabled.
However, the choice of `service.running`|`service.dead` here should also be a pillar; a pillar that `ci` and `prod` set to `True` but other environments leave to the `False` default.

giorgiosironi:
I see `is_prod_instance` has been removed from the builder, had a failure on `journal--end2end` when using the 3-days-old builder version. So I have to merge this to keep things straight; I'm favorable for it to be refactored to a pillar value.

PR #102, opened 2017-12-18
ELPP-3213 Retry Composer update

giorgiosironi:
```
[107.21.139.14] out: [ERROR   ] Command 'composer self-update' failed with return code: 1
[107.21.139.14] out: [ERROR   ] stderr: 
[107.21.139.14] out:   [Composer\Downloader\TransportException]
[107.21.139.14] out:   The "https://getcomposer.org/versions" file could not be downloaded: failed  to open stream: Connection timed out                                        
[107.21.139.14] out: self-update [-r|--rollback] [--clean-backups] [--no-progress] [--update-keys] [--stable] [--preview] [--snapshot] [--set-channel-only] [--] [<version>]
[107.21.139.14] out: [ERROR   ] retcode: 1
```

### DIFF
--- a/elife/scripts/update-composer.sh
+++ b/elife/scripts/update-composer.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+log_file="/tmp/update-composer-$(date -u +'%Y-%m-%dT%H:%M:%SZ').log"
+
+set -o pipefail
+if ! composer self-update 2>&1 | tee "$log_file"; then
+    echo "Composer update failure detected."
+    if grep "Connection timed out" "$log_file"; then
+        echo "Retrying update..."
+        composer self-update
+    else
+        echo "Failure does not seem retriable."
+        exit 2
+    fi
+fi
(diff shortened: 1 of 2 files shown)


elife/scripts/update-composer.sh
```diff
@@ -0,0 +1,16 @@
...
+log_file="/tmp/update-composer-$(date -u +'%Y-%m-%dT%H:%M:%SZ').log"
+
+set -o pipefail
+if ! composer self-update 2>&1 | tee "$log_file"; then
+    echo "Composer update failure detected."
+    if grep "Connection timed out" "$log_file"; then
+        echo "Retrying update..."
+        composer self-update
+    else
+        echo "Failure does not seem retriable."
```
thewilkybarkid: Is there any harm in just using the normal retry script though?

giorgiosironi: The original command fails after 60 seconds, if there are similar failure we should check them out rather than adding time to the build. In an ideal world this will be pushed down into a base image rather than being performed continuously.

thewilkybarkid: Ok, so to be expanded if we see more problems.

PR #103, opened 2017-12-19
Add coveralls placeholder

seanwiseman:


giorgiosironi:
It would make sense to have the `elife/coveralls.sls` state using this in the same PR too.

giorgiosironi:
https://github.com/elifesciences/elife-libraries-formula/blob/master/salt/elife-libraries/init.sls#L166-L174 needs to be copied and adapted into a new file `elife/coveralls.sls`.

giorgiosironi:
Then, `.coveralls` could be added at https://github.com/elifesciences/builder-base-formula/blob/master/elife/init.sls#L12

giorgiosironi:
Then (3rd step), add configuration to https://github.com/elifesciences/profiles-formula/blob/master/salt/pillar/profiles.sls#L31 and you should start seeing it in the `vagrant provision` result.

giorgiosironi:
The build is complaining because EC2 instances use `builder-private`, after adding the configuration there it should become green.

seanwiseman:
Looking good now @giorgiosironi ?

PR #107, opened 2018-01-19
ELPP-3361 Goaws configuration

giorgiosironi:
- create some queues by default
- specify an host for queues created by default

Not particularly useful in real environments, only for the Vagrant VM of `annotations` and other projects that now mimics `end2end` or `prod` much more than `ci`.


giorgiosironi:
Fails miserable on unrelated stuff like the JDK and MySQL PPA.

giorgiosironi:
Ok, it's just the JDK that fails and then cripples all subsequent `apt-get` commands.

giorgiosironi:
Now green.

PR #108, opened 2018-01-22
Update workaround for Java 8 PPA: 1.8.0_162

giorgiosironi:
Old workaround that fails on missing package from oracle.com needed some update to download the new package.

@lsh-0 you think we could try to move to openjdk in the long term?

### DIFF
--- a/elife/java8.sls
+++ b/elife/java8.sls
@@ -25,16 +25,16 @@ oracle-java8-installer:
 
     # temporary workaround because Oracle, licenses and the fact that the whole world software infrastructure is based on a single guy's free time
     # original work around, not working anymore: https://stackoverflow.com/questions/46815254/oracle-java8-installer-webupd8-ppa-404-not-found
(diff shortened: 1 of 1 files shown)


giorgiosironi:
`pr-base` should really have been failed:

It looks like it got the right package despite this change not being in https://alfred.elifesciences.org/blue/rest/organizations/jenkins/pipelines/pull-requests-projects/pipelines/builder-base-formula/branches/PR-108/runs/1/nodes/29/steps/32/log/?start=0
```
--2018-01-22 11:41:06--  http://download.oracle.com/otn-pub/java/jdk/8u161-b12/2f38c3b165be4555a1fa6e98c45e0808/jdk-8u161-linux-x64.tar.gz
```
So perhaps the package has been fixed upstream. I'm going to do an additional commit stripping out the workaround to see if that works.

giorgiosironi:
There has been a release in fact:
```
oracle-java8-installer | 8u161-1~webupd8~0 | Alin Andrei                                      (22 minutes ago)
-- | -- | --
```

PR #110, opened 2018-01-23
Feat customised updatedb

lsh-0:


elife/config/etc-updatedb.conf
```diff
@@ -0,0 +1,4 @@
+PRUNE_BIND_MOUNTS="yes"
+# PRUNENAMES=".git .bzr .hg .svn"
+PRUNEPATHS="/tmp /var/spool /media /home/.ecryptfs /ext"
+PRUNEFS="NFS nfs nfs4 rpc_pipefs afs binfmt_misc proc smbfs autofs iso9660 ncpfs coda devpts ftpfs devfs mfs shfs sysfs cifs lustre tmpfs usbfs udf fuse.glusterfs fuse.sshfs curlftpfs ecryptfs fusesmb devtmpfs"
```
giorgiosironi: because these filesystems are slow or shared, I guess

elife/config/etc-updatedb.conf
```diff
@@ -0,0 +1,4 @@
+PRUNE_BIND_MOUNTS="yes"
```
giorgiosironi: ok, I don't think we use bind mounts

elife/config/etc-updatedb.conf
```diff
@@ -0,0 +1,4 @@
...
+# PRUNENAMES=".git .bzr .hg .svn"
+PRUNEPATHS="/tmp /var/spool /media /home/.ecryptfs /ext"
```
giorgiosironi: do we use encryption anywhere? Well, no harm

elife/config/etc-updatedb.conf
lsh-0: or temporary or network shares

elife/config/etc-updatedb.conf
lsh-0: no, but I think builder-private is a real concern. I think new version of Salt allows you to mount an encrypted pillar on the master and target values within it

lsh-0:
that was the default config that came with ubuntu 14.04. I then did another commit with our elife-specific change `/ext` so it's clear what I changed in the blame history

PR #112, opened 2018-01-23
php5.6, doubled ram from 128 to 256.

lsh-0:
only crm uses php 5.6 currently with plans to upgrade to 7 soon

https://elifesciences.atlassian.net/browse/CIT-29


lsh-0:
4MB used to be pretty low. 32MB was pushing it. I think 128MB is the default on this 5.x version of php.

the problem this PR is for probably isn't a memory problem

PR #115, opened 2018-01-26
Simplify proofreader installation on VMs

giorgiosironi:
Abandon the Composer installation and install into a Git repository like for many other applications. From there it's a simple symlink into a folder on the PATH to use `proofreader`.

### DIFF
--- a/elife/proofreader-php.sls
+++ b/elife/proofreader-php.sls
@@ -1,6 +1,59 @@
-proofreader-php:
-    cmd.run:
-        - name: composer global --no-interaction require --update-with-dependencies elife/proofreader-php='dev-master#83901bb097b39b0df4ddc9d0e841b685977131ab'
-        - user: {{ pillar.elife.deploy_user.username }}
+proofreader-php-repository:
+    builder.git_latest:
+        - name: git@github.com:elifesciences/proofreader-php.git
+        - identity: {{ pillar.elife.projects_builder.key or '' }}
+        - rev: master
+        - branch: master
+        - target: /srv/proofreader-php
+        - force_fetch: True
+        - force_checkout: True
+        - force_reset: True
         - require:
             - composer
+
+    file.directory:
+        - name: /srv/proofreader-php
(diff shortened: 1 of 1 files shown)


elife/proofreader-php.sls
```diff
@@ -1,6 +1,59 @@
...
+        - recurse:
+            - user
+            - group
+        - require:
+            - builder: proofreader-php-repository
+
     cmd.run:
-        - name: composer global --no-interaction require --update-with-dependencies elife/proofreader-php='dev-master#83901bb097b39b0df4ddc9d0e841b685977131ab'
+        - name: |
+            composer install
```
thewilkybarkid: Since we want speed here, `--no-suggest --classmap-authoritative --no-dev`.

elife/proofreader-php.sls
```diff
@@ -1,6 +1,59 @@
...
         - user: {{ pillar.elife.deploy_user.username }}
+        - cwd: /srv/proofreader-php
```
thewilkybarkid: I'd always assumed `srv` = serve = web apps.

thewilkybarkid:
Should we not make use of https://github.com/elifesciences/proofreader-php/pull/5?

elife/proofreader-php.sls
giorgiosironi: Yes, this should never be used for development of `proofreader`.

elife/proofreader-php.sls
giorgiosironi: We install practically everything created after 2016 there, useful to sometimes mount `/srv` on an external volume and in general treating it as "where the `elife` user puts stuff"

giorgiosironi:
> Should we not make use of elifesciences/proofreader-php#5?

It would required `elife.docker` on all `*--ci` instances, which has implications for disk space usage and provisioning time. I'd move to it when each project moves towards containers (and hence its `--ci` instance disappears) while this should be backward compatible with its current `composer global` installation.

PR #118, opened 2018-01-29
ELPP-3361 Add Docker script for reading a label from any image and use it to provision sidecar containers

giorgiosironi:
Sidecar containers are used across applications to provide an instance of a common service.

We have several dummy implementation of services that can be used in this way, such as `api-dummy`.

Let's say I want to deploy `annotations` somewhere and I need to know which version of `api-dummy` it should be using:
```
$ elife/docker-scripts/docker-read-label elifesciences/annotations_cli 'org.elifesciences.dependencies.api-dummy' 2>/dev/null
05ab239fa0eec32b3f36ca2dc8117123e2879717
```

Once I know the version, I can run that and other sidecar images with its own `docker-compose` configuration:
```
version: '2'                                                                                                                                      
services:
    api_dummy:
        container_name: api_dummy
        image: elifesciences/api-dummy:${IMAGE_TAG}
        networks:
            - sidecars
        ports:
            # TODO: make target port customizable
            - 8001:8080
        restart: always

networks:
    sidecars:
        external: true
```

The new container joins a `sidecars` network that, when joined by an application too, makes it visible at `api-dummy:8080`. For ease of debugging, we make the sidecar also accessible on the host at `localhost:8001` if someone want to run a `curl` against it (the application doesn't rely on it instead).

### DIFF
--- a/_modules/elife.py
+++ b/_modules/elife.py
@@ -63,12 +63,14 @@ def image_tag():
         return 'latest' # conventionally stable Docker tag
     return revision
 
-def image_label(image, label):
-    """used to retrieve a `label` from an `image`.
+def image_label(image, label, image_tag='latest'):
+    """used to retrieve a `label` from an `image`. 
+    
+    Optionally specify an `image_tag` to pin.
 
     Warning: it will pull the image first, using bandwidth, time and disk space."""
 
-    return subprocess.check_output("/usr/local/docker-scripts/docker-read-label", image, label)
+    return subprocess.check_output("/usr/local/docker-scripts/docker-read-label", "%s:%s" % (image, image_tag), label)
 
 def read_json(path):
     "reads the json from the given `path`, detecting base64 encoded versions."
--- a/_modules/elife.py
+++ b/_modules/elife.py
@@ -78,6 +78,8 @@ def image_label(image, label, image_tag='latest'):
     value = output.strip()
     if value == 'null':
         raise RuntimeError("`%s` returned a null label" % command)
+    if value == '':
+        raise RuntimeError("`%s` returned a empty string label" % command)
     return value
 
 def read_json(path):
--- a/_modules/elife.py
+++ b/_modules/elife.py
@@ -1,5 +1,6 @@
 import os, json
 import base64
+import subprocess
 
 import logging
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


giorgiosironi:
The end game is that we move from specifying versions of dependencies in the `*.sha1` files like https://github.com/elifesciences/annotations/blob/develop/api-dummy.sha1 to these Docker image labels. That means we don't need to deploy the Git repository anymore on `end2end`/`continuumtest`/`prod` to read these values, but only the Docker image itself.

_modules/elife.py
```diff
@@ -62,6 +63,13 @@ def image_tag():
         return 'latest' # conventionally stable Docker tag
     return revision
 
+def image_label(image, label):
+    """used to retrieve a `label` from an `image`.
+
+    Warning: it will pull the image first, using bandwidth, time and disk space."""
```
giorgiosironi: This is a bit weird, but since we want to place the value of this label in a file, I couldn't find another way to execute some code before populating a `file.managed` template; it has to end up in a new environment variable here: https://github.com/elifesciences/annotations-formula/blob/master/salt/annotations/config/home-deployuser-annotations-.env

Fortunately, the image will be pulled anyway when it's actually used by the following states; so this is not adding time to the build as much as moving the operation up here instead (before any states are even executed I think, which is the misleading part).

elife/docker-scripts/docker-read-label
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 IMAGE LABEL"
+    echo "Example: $0 elifesciences/annotations_cli org.elifesciences.dependencies.api-dummy"
```
giorgiosironi: This is a bit incomplete because we should allow to specify a tag for the `elifesciences/annotations_cli` image. Better not to depend on what `latest` is this minute.

giorgiosironi:
Right now, dummies are configured like:
https://github.com/elifesciences/builder-private/blob/master/pillar/annotations.sls#L23-L26
but I envision something like:
```
        api_dummy:
            main_image: elifesciences/annotations_cli
            port: 8081 # 8082 for https
```
using the tag from build vars to pull a particular version of `main_image`, and  the standard `org.elifesciences.dependencies.$pillarKey` for the label. The port may still be useful to configure for now.

giorgiosironi:
I'd go for another round of review since all the Salt part was added today.

lsh-0:
:+1:

PR #122, opened 2018-02-12
default psql permissions for non-legacy databases

lsh-0:


elife/postgresql-appdb.sls
```diff
@@ -20,7 +20,7 @@
 db-perms-to-rds_superuser:
 {% if db_exists and app_user_exists %}
```
giorgiosironi: I'm not sure about how one or the other script is chosen to run. Is `*-legacy.sh` supposed to run every time after the first provisioning (where `-fresh.sh` would run)? Or is it only for particular existing RDS databases? But how can the latter case be detected?

lsh-0: yes, it's logic is a little hard (open to suggestions).

when this change was introduced, an existing db would have been detected and the legacy script run. that script shifts permissions to the root user and makes sure the app user can do what they need to do. it worked well.

on new instances that legacy script isn't being run because no db is being detected. previously there was just a short message "no database to wrangle". what it does now with new databases is assign some more permissions to the root user. this is only necessary when using RDS because the pseudo-root user doesn't have permission to deal with new objects in a database, only the ones that were there at the time the permissions were assigned - which was just the empty database. What this script does is assign the pseudo-root user permissions on current objects and _all future objects_ too.

lsh-0: hrm - I think I see your problem actually. the legacy script will be run on non-legacy databases, but only once and only if the flag doesn't exist. if both states use the same flag we should be fine, but this logic _is_ getting convoluted.

lsh-0: I've tweaked the scripts so both are setting the same flag which reduces the complexity of logic a little.

elife/postgresql-appdb.sls
```diff
@@ -18,10 +18,13 @@
 
 # handles permissions on legacy databases
 db-perms-to-rds_superuser:
-{% if db_exists and app_user_exists %}
     cmd.script:
-        - name: salt://elife/scripts/postgresql-appdb-perms.sh
-        - creates: /root/legacy-db-permissions-migrated.flag
+    {% if db_exists and app_user_exists %}
+        - name: salt://elife/scripts/postgresql-appdb-perms-legacy.sh
```
giorgiosironi: so if I understand correctly this terminology
1. `legacy` should run once on all stacks, then we can remove it?
2. We will then leave only `fresh` which will only be run once on the first successful high state, then never again?
3. If the above is true, can we add `TODO` comments saying when to remove `legacy`?

giorgiosironi:
Evaluate whether we should change https://github.com/elifesciences/tech-team/blob/master/adr/0003-rds-postgresql-permissions.md to `superseded` and add a new one (probably very similar but with crucial differences.)

elife/postgresql-appdb.sls
lsh-0: all projects that opted in to the postgresql-appdb state should have had the legacy script run already. future projects that decide to opt in will need that script run on them to migrate any permissions. until the appdb work becomes default, and I don't think it should, the legacy script must remain.

giorgiosironi: > until the appdb work becomes default, and I don't think it should, the legacy script must remain.

So can it be named something like `migration-to-appdb` or similar, rather than `legacy` which as a name pushes for its own cleanup?

lsh-0: ah, I see your problem now. I wasn't thinking of legacy as legacy. I was thinking of legacy as legacy. http://www.dictionary.com/browse/legacy

I'll rename it so it's clearer

lsh-0:
I just re-read that ADR and it's still good but would tweak the permissions description of the application user to be able to modify the objects within the database that they created. The database essentially becomes the application user's sandbox.

PR #125, opened 2018-02-21
ELPP-3476 Comment out New Relic Infrastructure installation to mitigate bad repository

giorgiosironi:
Existing instances should continue to run the daemon, but new one won't until restored.


giorgiosironi:
Problems seems resolved upstream so closing for now.

PR #127, opened 2018-02-22
Add elife.postgresql-container

giorgiosironi:
For local usage, it seems easy enough now to use the `postgres` official Docker image rather than installing it through the apt repositories:

- it's a charm to create a root user as it can be specified through environment variables (not very safe, but production uses RDS anyway)
- There is stronger isolation as the filesystem is different from the host's one, so this is more similar to an external host like RDS than a local service.
- In testing, cleaning the database should be as simple as `docker-compose down` rather than querying here and there and recreating objects with the wrong permissions every time.

All interaction has to happen through `localhost:5432` that is forwarded
to the container's port. Similarly to `sidecars`, the container joins a
network called `databases` that application containers have to opt-in to
access.


giorgiosironi:
Note `elife.postgresql-appdb` should in theory work transparently with this, which is alternative to `elife.postgresql`.

lsh-0:
not an alternative, `postgresql-appdb` requires `postgresql`

PR #133, opened 2018-03-12
ELPP-3432 Support PyPI credentials

giorgiosironi:
Trying a different approach to the usual: to achieve better segregation, putting the credentials into a separate `pypi` pillar. This should only be included on the instances that actually need it rather than everywhere like for the blanket `pillar.elife.*`.

### DIFF
--- a/elife/config/home-deploy-user-.pypirc
+++ b/elife/config/home-deploy-user-.pypirc
@@ -1,0 +1,5 @@
+[pypirc]
+servers = pypi
+[server-login]
+username: {{ pillar.pypi.username }}
+password: {{ pillar.pypi.password }}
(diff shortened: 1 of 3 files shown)


giorgiosironi:
See the referencing PR for an example.

giorgiosironi:
>  it's a sign we're probably not doing security as well as we could.

We could store credentials in Jenkins so that they can be passed only to the steps that need them - but it's a big lock in so I prefer infrastructure to reusable outside of Jenkins.

> are the credentials to push a package to pypi automatically .. ?

Yes https://elifesciences.atlassian.net/browse/ELPP-3432

lsh-0:
I was thinking encrypted values in the pillar data or something like Hashicorp's Vault.

ty for link, I added a DOD

PR #136, opened 2018-03-13
ELPP-3163 Allow specific version of proofreader

giorgiosironi:
For a more isolated rollout.

### DIFF
--- a/elife/proofreader-php.sls
+++ b/elife/proofreader-php.sls
@@ -2,7 +2,7 @@ proofreader-php-repository:
     builder.git_latest:
         - name: git@github.com:elifesciences/proofreader-php.git
         - identity: {{ pillar.elife.projects_builder.key or '' }}
-        - rev: 0.1
+        - rev: {{ pillar.elife.proofreader_php.version }}
         - branch: master
         - target: /srv/proofreader-php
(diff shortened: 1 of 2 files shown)


elife/proofreader-php.sls
```diff
@@ -23,7 +23,7 @@ proofreader-php-repository:
 
     cmd.run:
         - name: |
-            composer --no-interaction install --no-suggest --classmap-authoritative --no-dev
+            composer --no-interaction update --no-suggest --classmap-authoritative --no-dev
```
giorgiosironi: This can probably be reverted after the upgrade for speed, until we remove this state file at all due to containerization.

pillar/elife.sls
```diff
@@ -183,6 +183,8 @@ elife:
                 #    command: /srv/annotations/bin/console queue:watch
                 #    number: 1
                 #    [require: some-state]
+    proofreader_php:
+        version: 0.1
```
thewilkybarkid: Since we'll mostly just be adding rules rather than changing them, we can update the code first then bump this, rather than having to declare the version in the repo.

giorgiosironi: Interesting, let's test that out with medium.

PR #139, opened 2018-04-04
Only give credentials to www-data if user exists

giorgiosironi:
The user always exists, so this doesn't work. But should we always give it credentials?

### DIFF
--- a/elife/aws-credentials.sls
+++ b/elife/aws-credentials.sls
@@ -22,4 +22,7 @@ aws-credentials-www-data-user:
         - group: {{ pillar.elife.webserver.username }}
         - makedirs: True
         - template: jinja
+        - onlyif:
+            # if user exists
+            - id www-data
         # no clear require, since webserver user could be created by uwsgi, php-fpm, nginx...
--- a/elife/aws-credentials.sls
+++ b/elife/aws-credentials.sls
@@ -23,6 +23,6 @@ aws-credentials-www-data-user:
         - makedirs: True
         - template: jinja
         - onlyif:
-            # if user exists
-            - id www-data
+            # if user home folder exists
+            - test -d /var/www
         # no clear require, since webserver user could be created by uwsgi, php-fpm, nginx...


giorgiosironi:
`basebox` has it:
```
[09:03:26][elife@dev--basebox:~]$ id www-data
uid=33(www-data) gid=33(www-data) groups=33(www-data)
```
but it shouldn't install anything like `nginx` or `apache`. Closing for now, don't see an easy solution; at least the state will never fail which was the initial thing that worried me.

thewilkybarkid:
IIRC the user's always there, but not its home directory.

giorgiosironi:
`/var/www` is indeed missing:
```
[10:41:54][elife@dev--basebox:~]$ grep -r www-data /etc/passwd
www-data:x:33:33:www-data:/var/www:/usr/sbin/nologin
[10:41:57][elife@dev--basebox:~]$ ls -l /var
total 44
drwxr-xr-x  2 root root   4096 Apr 10  2014 backups
drwxr-xr-x 12 root root   4096 Apr  5 08:56 cache
drwxr-xr-x  3 root root   4096 Apr  5 08:54 chef
drwxrwxrwt  2 root root   4096 Jun 23  2016 crash
drwxr-xr-x 49 root root   4096 Apr  5 08:58 lib
drwxrwsr-x  2 root staff  4096 Apr 10  2014 local
lrwxrwxrwx  1 root root      9 Jun 23  2016 lock -> /run/lock
drwxrwxr-x 11 root syslog 4096 Apr  5 08:58 log
drwxrwsr-x  2 root mail   4096 Jun 23  2016 mail
drwxr-xr-x  2 root root   4096 Jun 23  2016 opt
lrwxrwxrwx  1 root root      4 Jun 23  2016 run -> /run
drwxr-xr-x  4 root root   4096 Apr  5 08:58 spool
drwxrwxrwt  2 root root   4096 Jun 23  2016 tmp
```

PR #140, opened 2018-04-04
Automated resize of filesystem on external volume

giorgiosironi:
If we expand a volume through CloudFormation (or the EBS API in
general), the filesystem on it needs to be expanded to take advantage of
the newly available space.

This ensures a Salt highstate will make `df -h` see all the space that
is not available through `lsblk`.


giorgiosironi:
> The size parameter

Fortunately it appears to work it out by itself if not passed :+1:

PR #141, opened 2018-04-10
known-hosts, fixes bug in Salt upgrade where default hash_type changed

lsh-0:
see:
* https://docs.saltstack.com/en/2017.7/ref/states/all/salt.states.ssh_known_hosts.html#salt.states.ssh_known_hosts.present
* https://docs.saltstack.com/en/2017.7/ref/modules/all/salt.modules.ssh.html#salt.modules.ssh.recv_known_host

you can do this to play around with the parameters `sudo salt-call ssh.recv_known_host bitbucket.org enc=ssh-rsa`

### DIFF
--- a/elife/known-hosts.sls
+++ b/elife/known-hosts.sls
@@ -6,7 +6,9 @@ github.com:
     ssh_known_hosts.present:
         - fingerprint: 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48
         - enc: ssh-rsa
+        - fingerprint_hash_type: md5
         - unless:
+            # BUG: doesn't seem to be working, entries are accumulating
             - grep -r "^github.com," /etc/ssh/ssh_known_hosts
 
 


giorgiosironi:
Older salt seems not to like this, or at least it does produce a warning:
```
[54.210.181.244] out:     Warnings: 'fingerprint_hash_type' is an invalid keyword argument for
[54.210.181.244] out:               'ssh_known_hosts.present'. If you were trying to pass additional
[54.210.181.244] out:               data to be used in a template context, please populate 'context'
[54.210.181.244] out:               with 'key: value' pairs. Your approach will work until Salt Oxygen
[54.210.181.244] out:               is out. Please update your state files.
```
The other failure is 
```
buildercore.config.ConfigurationError: a FASTLY_API_KEY environment variable is required to provision Fastly resources
```
which I thought would have been skipped if no Fastly is needed. I'll make sure it is.

lsh-0:
it's just a warning in older salt, which won't be there after the migration

I fixed your other problem, but I'm still going through my emails

PR #147, opened 2018-06-14
ELPP-3693 Allow to run builds as the ubuntu user

giorgiosironi:
This node is used by Jenkins to run builds or part of builds. It usually uses SSH access through the `elife` user, but for simplicity the EC2 plugin should use the `ubuntu` user instead.  

The plugin creates throwaway instances from an AMI, hence its own configured keypair is only added to the `ubuntu` user and `elife` is locked out (all access is removed in the AMI creation process). Human users can still log in through the `elife-alfred--prod` instance, which has the keypair.

### DIFF
--- a/elife/docker.sls
+++ b/elife/docker.sls
@@ -71,11 +71,12 @@ docker-compose:
         - require:
             - file: docker-compose
 
-docker-deploy-user-in-group:
+docker-users-in-group:
     group.present:
         - name: docker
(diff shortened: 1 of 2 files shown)


giorgiosironi:
The disk on heavybox seems full. Perfect timing for https://github.com/elifesciences/builder/pull/347 which will allow larger root volumes.

PR #151, opened 2018-07-04
removes the infrastructure client and kills the client application

lsh-0:


elife/newrelic-infrastructure-removal.sls
```diff
@@ -0,0 +1,10 @@
+newrelic-infrastructure-configuration:
+    file.absent:
+        - name: /etc/newrelic-infra.yml
+
+newrelic-infrastructure:
+    pkg.purged:
+        - name: newrelic-infra
+
+    service.dead:
```
giorgiosironi: maybe do first `service.dead` and then `pkg.purged`? Although `pkg.purged` may cover everything already

lsh-0:
I wouldn't have thought it mattered but I think I've been on arch+systemd for too long. Installing a service and starting/stopping it are separate. apt ... unsure. I'll switch it's order.

giorgiosironi:
The output of the build looks good, no permission denied errors.

lsh-0:
it all went well, thanks for the builder fix

PR #152, opened 2018-07-05
Avoid cron emails

giorgiosironi:
By default cron sends emails (with the output of commands that are run)
to local users, who will never read the content e.g.
`elife@prod--journal-cms.elifesciences.org`.

By setting the environment in this way, emails should not be created.


giorgiosironi:
Local tests show a 1-minute cron running in `/var/log/cron` but an empty `/var/mail/`.

PR #153, opened 2018-07-16
Make set_local_revision work on AWS

giorgiosironi:
Build vars are environment information kept in a JSON document encoded in base 64, first provisioned upon an EC2 instance creation.

The current script overwrites the whole file, which works in Vagrant but deletes important information like [RDS usernames and passwords](https://github.com/elifesciences/journal-cms-formula/blob/master/salt/journal-cms/init.sls#L146).

### DIFF
--- a/elife/utils/set_local_revision
+++ b/elife/utils/set_local_revision
@@ -2,6 +2,7 @@
 set -e
 
 if [ "$#" -ne "1" ]; then
+    echo "Set the branch or commit of the project to be used for the next Salt highstate on this node. Works in both Vagrant and AWS"
     echo "Usage: set_local_revision BRANCH|COMMIT"
     echo "Example: set_local_revision develop"
     echo "Example: set_local_revision 964a28d97275ceebfe15e87b8ce86f90228ae3a0"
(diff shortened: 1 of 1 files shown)


elife/utils/set_local_revision
```diff
@@ -2,11 +2,25 @@
...
+    echo '{}' | base64 | sudo tee "${build_vars}"
+fi
+
+# preserve any existing build vars like `rds_username` to allow usage on AWS
+cat "${build_vars}" \
+    | base64 -d \
+    | jq ".revision = \"$revision\"" \
+    | base64 \
+    | sudo tee "${build_vars}.new"
+sudo mv "${build_vars}.new" "${build_vars}"
```
thewilkybarkid: Is the intermediate file needed?

giorgiosironi: It's all pipes so could run into some obscure runs where the input file is written to before it is completely read

thewilkybarkid:
Have now tried this out, and it works.

It does put the base64-encoded string on stdout, which isn't that friendly but is harmless.

PR #155, opened 2018-07-18
installation of python/pip/setuptools is now handled in bootstrap script

lsh-0:
removes a lot of cruft for an ancient 'fix'

(depends on https://github.com/elifesciences/builder/pull/365)

### DIFF
--- a/elife/python.sls
+++ b/elife/python.sls
@@ -75,6 +75,10 @@ python-dev:
             - libffi-dev 
             - libssl-dev
 
+python-pip:
+    cmd.run:
+        - name: echo Managed by builder. Remove when no other formulas depend on it
+
 global-python-requisites:
     cmd.run:
         # rmrf_enter is DEPRECATED
--- a/elife/python.sls
+++ b/elife/python.sls
@@ -66,14 +66,7 @@ global-python-requisites:
 
 # DEPRECATED. removed after switch to 16.04
 
-python-2.7:
-    pkg.installed:
-        - name: python2.7
-        # these don't work to upgrade if 'python2.7' already installed
-        # and it's in the base box so this should never do anything
-        #- refresh: True
-        #- allow_updates: True
-        #- reload_modules: True
+# WARN: assumes python 2.7.12+ are already installed via bootstrap script
 
 python-dev:
     pkg.installed:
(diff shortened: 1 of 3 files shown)


elife/python.sls
```diff
@@ -66,14 +66,7 @@ global-python-requisites:
 
 # DEPRECATED. removed after switch to 16.04
```
giorgiosironi: does the 16.04 part need any change given the shift of the installation into builder?

lsh-0: probably, I didn't look at it.

giorgiosironi: The 16.04 all seems backward compatible, but cross-check please

giorgiosironi:
Several states in elife-bot-formula are using `requires` on these:

```
require:
    elife-bot--vagrant:                                      pip: global-python-requisites
 require:
    elife-bot--vagrant:                                      pkg: python-pip
```

I'll add them as empty states for backward compatibility.

giorgiosironi:
They are also too specific (`pip`, `pkg`) so I'll have to amend the formula anyway first.

giorgiosironi:
Java installation is failing and wrecking apt-get. Usually when Oracle removes old archives this happens.

giorgiosironi:
Ok, the Java installation is a recurring problem that happens every time Oracle removes the old versions of binaries. Here is some links to work around it @lsh-0 if you happen to be around, don't know your schedule.

The stuff to uncomment: https://github.com/elifesciences/builder-base-formula/blob/master/elife/java8.sls#L31-L37
What to put in for the numbers (untested): 
- https://twitter.com/behemphi/status/953372016677146630
- https://twitter.com/behemphi/status/953371839937630208


(and then we move to OpenJDK or containers down the road?)

lsh-0:
hey @giorgiosironi , yes I'm about today but not tomorrow. I'll take a look at this java stuff.

I think elife-bot is the only one still using rmrf_enter as well, an opportunity there to remove some complexity. I didn't touch it this time around because `pip.installed` should be working regardless of what it's installing, but food for thought.

lsh-0:
the java stuff appears to work fine on fresh instances (tested on vagrant) and subsequent highstates, but I had a problem with dpkg needing to be reconfigured on existing instances. I'll push my changes and see if we can replicate, but worst case we have to do some manual wrangling on java enabled instances

lsh-0:
@giorgiosironi , I'm not going to push any changes up regarding the java stuff, it appears to be working now. I can build a fresh an instance with it as-is. The dpkg problems I was having before may have been due to uncommenting all that `sed` stuff.

lsh-0:
four failed states, all unrelated to java. 

> I'll add them as empty states for backward compatibility.

these failed because these states weren't found. I'll add them in now.

lsh-0:
looking good! merge if you're happy

PR #156, opened 2018-07-19
Kubectl, gcloud and other tooling for deploying to GKE

giorgiosironi:
The tool of choice for interacting with Kubernetes clusters. This should be managed by builder in the long term, but unless it provides within a container, the binary still has to be present.

The configuration contains sensitive credentials to access the cluster.

### DIFF
--- a/elife/gcloud.sls
+++ b/elife/gcloud.sls
@@ -1,0 +1,20 @@
+google-cloud-packages-repo:
+    cmd.run:
+        - name: curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+        - unless:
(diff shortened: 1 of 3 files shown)


giorgiosironi:
We need to login with `gcloud` rather than just using `kubectl`, then we can use
```
gcloud container clusters get-credentials data-pipeline-cluster --zone us-east4-a                                             
```
to generate `.kube/config`. The file to store is the private key for the Service Account which is more general than the kubectl configuration.

lsh-0:
I understand not wanting to be blocked by code review but if you could @ping me on significant PRs I'd be grateful.

lsh-0:
... and to whomever @ping is, sorry for the spam.

PR #159, opened 2018-07-23
Extract uwsgi params

giorgiosironi:
This file can be used by an nginx on the host pointing to a container, so it doesn't need to be tied to installing uwsgi on the host.

It's probably generic enough to be called `wsgi-params` referring to the protocol rather than the daemon; however renaming requires other operations involving nginx so it can't be done at this time.

 Also necessary for https://github.com/elifesciences/jira-import/issues/4292


giorgiosironi:
> restructuring builder-base to make it more container friendly

In an ideal world wouldn't be necessary to use it at all as we would deploy containers on an orchestrator (which deals with configuration with a separate repository), but due to the transition period we need some EC2 nodes daemons like `nginx` to collaborate with containers running on them

PR #161, opened 2018-07-31
filters bootstrap user's authorized keys when masterless

lsh-0:
this should prevent the scenario I'm dealing with right now

### DIFF
--- a/elife/masterless.sls
+++ b/elife/masterless.sls
@@ -11,10 +11,16 @@ disable-salt-minion:
 # deny access to all except those in allow.all and allow.master-server
 # necessary to ensure those with access don't inadvertently grant access to others
 
+{% set stackname = salt['elife.cfg']('project.stackname') | replace('-', '\-') %}
+
 # zero out all allowed keys
 deny_all_access:
     cmd.run:
-        - name: echo > /home/elife/.ssh/authorized_keys
+        - name: |
+            set -e
+            echo > /home/elife/.ssh/authorized_keys
+            cat /home/ubuntu/.ssh/authorized_keys | grep "{{ stackname }}" > tmp
+            mv tmp /home/ubuntu/.ssh/authorized_keys
         - require:
             - ssh-access-set # regular permissions configured
 
--- a/elife/masterless.sls
+++ b/elife/masterless.sls
@@ -11,15 +11,13 @@ disable-salt-minion:
 # deny access to all except those in allow.all and allow.master-server
 # necessary to ensure those with access don't inadvertently grant access to others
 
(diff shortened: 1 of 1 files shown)


lsh-0:
It's failing here:
```
[54.82.239.211] out:                   Collecting spectrumprivate from git+ssh://git@github.com/elifesciences/elife-spectrum-private.git@91371987bee96747b9cef6dfb7f63d0b04d4a048#egg=spectrumprivate (from -r requirements.txt (line 38))
[54.82.239.211] out:                     Cloning ssh://git@github.com/elifesciences/elife-spectrum-private.git (to revision 91371987bee96747b9cef6dfb7f63d0b04d4a048) to /tmp/pip-install-MM1tzz/spectrumprivate
[54.82.239.211] out:                   Warning: Permanently added the RSA host key for IP address '192.30.253.112' to the list of known hosts.
[54.82.239.211] out:                   Permission denied (publickey).
[54.82.239.211] out:                   fatal: Could not read from remote repository.
[54.82.239.211] out:                   
[54.82.239.211] out:                   Please make sure you have the correct access rights
[54.82.239.211] out:                   and the repository exists.
```

do you have any insight into this @giorgiosironi ? I wouldn't have thought it had anything to do with `authorized_keys` file

elife/masterless.sls
```diff
@@ -14,7 +14,11 @@ disable-salt-minion:
 # zero out all allowed keys
 deny_all_access:
     cmd.run:
-        - name: echo > /home/elife/.ssh/authorized_keys
+        - name: |
+            set -e
+            echo > /home/elife/.ssh/authorized_keys
+            cat /home/ubuntu/.ssh/authorized_keys | grep -E "{{ salt['elife.cfg']('project.stackname') }}|basebox" > tmp
```
giorgiosironi: not sure what the `grep` is preserving?

lsh-0: ah, there are two keys in /home/ubuntu/.ssh/authorized_keys that should probably be preserved. a pubkey with the name of the stack, like `lax--prod`, probably put there automatically by the ami/ec2 creation system so we can login. The other is a key called 'basebox' that I'm not sure about - I included it thinking it might solve the build error we're seeing here but that wasn't the case. The 'basebox' key might be the pubkey from the original ami, in which case it could probably be removed, but I thought you might be depending on it for some reason.

lsh-0:
thanks for merging the fix

lsh-0:
I can confirm the bootstap user's authorized_keys file is now correctly truncated - at least on this masterless peerscout instance I just brought up

PR #164, opened 2018-08-06
Script waiting for Docker daemon to start

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4377

We have had builds like https://alfred.elifesciences.org/job/test/job/test-digests/22/console:
```
[1748] Failed to execute script docker-compose
...
requests.exceptions.ReadTimeout: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
```

The Docker daemon taking >1 minute to start seems to be the culprit, the underlying cause probably being the fact that this may be a brand new EC2 instance or having just been rebooted, hence needing to populate caches and do general bootstrap work.

This script should help to wait for the daemon to be available before starting builds that will fail in this way.

### DIFF
--- a/elife/docker-scripts/docker-wait-daemon
+++ b/elife/docker-scripts/docker-wait-daemon
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+timeout="${1:-120}"
+socket="/var/run/docker.sock"
+
+echo "Waiting for socket $socket to be writable"
+# similar to elife/utils/wait_for_port but with Unix sockets (-U)
+timeout "$timeout" sh -c "while ! nc -q0 -w1 -U $socket </dev/null >/dev/null 2>&1; do sleep 1; done"
+echo "Connection was established on $socket"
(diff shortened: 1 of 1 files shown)


elife/docker-scripts/docker-wait-daemon
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+timeout="${1:-120}"
+socket="/var/run/docker.sock"
+
+echo "Waiting for socket $socket to be writable"
+# similar to elife/utils/wait_for_port but with Unix sockets (-U)
+timeout "$timeout" sh -c "while ! nc -q0 -w1 -U $socket </dev/null >/dev/null 2>&1; do sleep 1; done"
```
lsh-0: I use `timeout` in strip-coverletter to prevent excessively long squashes. Good stuff. I would add a comment beneath your comment though explaining this soupy bash code. Something like

> poll docker socket once a second until timeout is reached (default 120 seconds)

and if `nc` has long forms for those parameters (`-q`, `-w`, `-U`), use those.

giorgiosironi: Unfortunately, no long form

PR #166, opened 2018-08-08
Mysql 5.7 update

giorgiosironi:
Old version has been removed from repositories, hence we need to update to fix automated builds and highstates on new instances


giorgiosironi:
Merge if satisfied, prevents build errors unrelated from their proposed change.

giorgiosironi:
I would have expected `pr-base` to fail rather than `pr-update`, checking the logs.

giorgiosironi:
`pr-update` error is unrelated, needs `master` to be merged in.

giorgiosironi:
Ah, the state is not run in [`heavybox`](https://github.com/elifesciences/heavybox-formula/blob/master/salt/example.top#L34), fair enough, `pr-base` won't catch it. Merging master and all should be green, albeit not testing that much.

PR #169, opened 2018-08-20
adds logrotate rules for ubr

lsh-0:
Brought a vagrant machine up for this and it worked well. Logrotate is often fussy with dir permissions but I made sure it was being rotated correctly.

### DIFF
--- a/elife/backups.sls
+++ b/elife/backups.sls
@@ -44,3 +44,7 @@ monitor-backup-logs:
         - require:
             - pkg: syslog-ng
 
+rotate-backup-logs:
+    file.managed:
+        - name: /etc/logrotate.d/ubr
+        - source: salt://elife/config/etc-logrotate.d-ubr
--- a/elife/config/etc-logrotate.d-ubr
+++ b/elife/config/etc-logrotate.d-ubr
@@ -1,0 +1,8 @@
+/opt/ubr/ubr.log {
+    daily
+    rotate 7
+    notifempty
+    missingok
+    compress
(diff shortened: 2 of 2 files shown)


elife/backups.sls
```diff
@@ -44,3 +44,7 @@ monitor-backup-logs:
         - require:
             - pkg: syslog-ng
 
+rotate-backup-logs:
+    file.managed:
+        - name: /etc/logrotate.d/ubr
+        - source: salt://elife/config/etc-logrotate.d-ubr
```
giorgiosironi: do you need to trigger in someway `logrotate` to re-read this? I don't think so as it's not a daemon, and should read it when started by `cron`

elife/config/etc-logrotate.d-ubr
```diff
@@ -0,0 +1,8 @@
+/opt/ubr/ubr.log {
+    daily
+    rotate 7
+    notifempty
+    missingok
+    compress
+    copytruncate
```
giorgiosironi: `copytruncate` is usually for daemons as the file is emptied in place and the daemon can continue writing to it. It does no harm though

lsh-0: (I copied+pasted from another logrotate file)

elife/backups.sls
lsh-0: I don't think so. logrotate, like most shell things, will pick up on the contents of it's `.d/` directory

PR #171, opened 2018-08-28
basebox 18.04 support

giorgiosironi:
@lsh-0 add your own packages too.

### DIFF
--- a/elife/python.sls
+++ b/elife/python.sls
@@ -34,12 +34,12 @@ python-2.7:
         - require:
             - third-party-python-repos-absent
 
-python-3.5:
+python-3:
     pkg.installed:
         - pkgs:
-            - python3.5
+            - python3
             - python3-pip
-            - python3.5-venv
+            - python3-venv
         - require:
             - third-party-python-repos-absent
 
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Missing `realpath` as [Java 7 has been removed](https://github.com/elifesciences/builder-private/commit/dc3c9695ea15db7a9707e626dce02e504f3514d7)

giorgiosironi:
There are all sort of checks for `16.04` or `trusty` around the various states, I'm limiting myself to what is necessary for `basebox` to be built for now.

elife/python.sls
```diff
@@ -1,4 +1,5 @@
-{% if salt['grains.get']('osrelease') == "16.04" %}
+{% set osrelease = salt['grains.get']('osrelease') %}
+{% if osrelease == "16.04" or osrelease == "18.04" %}
```
lsh-0: perhaps `{% if osrelease != "14.04" %}` which is a bit more future proof

giorgiosironi:
A Vagrant VM with `bento/ubuntu-18.04` builds correctly, but this build will run more stuff in heavybox so it will probably still fail.

giorgiosironi:
Ok, this passes, but only because it's built on 14.04 for now. Renaming to `basebox support` and merging as a first step.

PR #173, opened 2018-08-28
websocket support

damnhipster:
`elife-xpub` is using WebSockets to report on the progress of file uploads. This change allows http and https connections to be upgraded to `ws://` and `wss://`.


damnhipster:
Addressed on https://github.com/elifesciences/elife-xpub-formula

giorgiosironi:
Good call, this is used on so many project that it's easier to roll it out for xpub only.

PR #177, opened 2018-09-11
Revert "Revert "Feature: add python3.6-venv to list of packages installed during bootstrap""

lsh-0:
Reverts elifesciences/builder-base-formula#176

FML it *is* required as it's failing without it in end2end. I hate ubuntu *so much* sometimes.


giorgiosironi:
Would be helpful to know what is failing without it (which project)?

PR #178, opened 2018-09-14
logging, adds custom daily logrotate script to log any errors

lsh-0:
the modified part of the script: https://github.com/elifesciences/builder-base-formula/blob/7ae91be3f971faea040a3db16ab4b2388f85fb96/elife/config/etc-cron.daily-logrotate#L14-L19


lsh-0:
I just ran `run-parts /etc/cron.daily` and the output was sent to the terminal rather than the capture file. grr

elife/config/etc-cron.daily-logrotate
```diff
@@ -0,0 +1,19 @@
...
+head -1 status > status.clean
+sed 's/"//g' status | while read logfile date
+do
+    [ -e "$logfile" ] && echo "\"$logfile\" $date"
+done >> status.clean
+mv status.clean status
+
+test -x /usr/sbin/logrotate || exit 0
+capture_file=/tmp/daily-logrotate.out
+/usr/sbin/logrotate /etc/logrotate.conf --verbose &> "$capture_file" || {
```
giorgiosironi: `&>` should redirect both stdout and stderr, but I've usually seen it without spaces between operator and file name.

lsh-0: like this? ` --verbose &>"$capture_file" ` 

I can't say I've seen it consistently done one way or the other. 

I think cron may be futzing with the stdout/stderr streams anyway, I need to do some more investigation

elife/logrotate.sls
```diff
@@ -0,0 +1,5 @@
+custom daily logrotate script:
+    file.managed:
+        - name: /etc/cron.daily/logrotate
+        - source: salt://elife/config/etc-cron.daily-logrotate
+        - mode: 755
```
giorgiosironi: does this need a `require` to be done after `logrotate` installation, which would presumably also create this cron on new instances?

elife/config/etc-cron.daily-logrotate
giorgiosironi: I confirm it works in all expected cases :+1:

elife/logrotate.sls
lsh-0: not sure, wouldn't have thought so but I've added it anyway

giorgiosironi:
Error on `smoke.sh`, I'll merge `master` into here to use the latest version.

lsh-0:
thanks, @giorgiosironi . There is still a problem with logrotate not redirecting stdout/err that needs to be figured out before I merge this.

lsh-0:
found it and fixed it. problem was with `sh` redirections vs `bash`. also, logrotate doesn't return a non-zero status when there is a problem so some grepping was involved.

lsh-0:
merging

lsh-0:
doh

PR #179, opened 2018-09-21
Only print pinned commit in update dependency pipeline

giorgiosironi:
Titles like:

```
Updated digestparser: pinned digestparser in requirements.txt to 88c712a4d4d9c4a45fefbb820859151de49d8cc9
```

can be reduced to:

```
Updated digestparser: 88c712a4d4d9c4a45fefbb820859151de49d8cc9
```

Sample from https://github.com/elifesciences/elife-bot/pull/779


giorgiosironi:
A problem on 18.04 for the build in general, not sure it that will reproduce as it looks like a race condition with a cron.

PR #181, opened 2018-10-09
Allow customization of last hours limit in docker-prune

giorgiosironi:


giorgiosironi:
Temporary (I think) failure on 18.04 version.

PR #184, opened 2018-10-18
General support for adding SSH credentials

giorgiosironi:


giorgiosironi:
18.04 continues to be broken.

PR #185, opened 2018-10-18
Run docker-prune on each highstate

giorgiosironi:
Trialled in `elife-xpub`, works well and we go for a reasonable default of 14 days to avoid losing caches. Can be tuned per-project.

### DIFF
--- a/elife/docker.sls
+++ b/elife/docker.sls
@@ -103,3 +103,11 @@ docker-ready:
             - docker-users-in-group
             - docker-scripts
             - docker-scripts-path
+
+# frees disk space from old images/containers/volumes/...
+# older than last X days hours and not in use
+docker-prune-last-days:
+    cmd.run:
+        - name: /usr/local/docker-scripts/docker-prune {{ 24 * pillar.elife.docker.prune_days }}
+        - require:
+            - docker-ready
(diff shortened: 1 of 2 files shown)


elife/docker.sls
```diff
@@ -103,3 +103,11 @@ docker-ready:
...
             - docker-scripts
             - docker-scripts-path
+
+# frees disk space from old images/containers/volumes/...
+# older than last X days hours and not in use
+docker-prune-last-days:
+    cmd.run:
+        - name: /usr/local/docker-scripts/docker-prune {{ 24 * pillar.elife.docker.prune_days }}
+        - require:
+            - docker-ready
```
thewilkybarkid: Probably doesn't matter when this happens.

giorgiosironi: Must happen after the `docker` binary is available e.g. on installation (in that case it would be almost a no-op)

giorgiosironi:
Have to debug this tomorrow, some pillars not correctly set.

elife/docker.sls
giorgiosironi: No, it should not matter when it happens; `docker-ready` is mostly for newly created or rebooted VMs so in that case they shouldn't have an almost full disk.

giorgiosironi:
Magically also 18.04 worked?

PR #186, opened 2018-10-25
Remove pip.installed

giorgiosironi:
From https://github.com/elifesciences/elife-xpub-formula/pull/51 `pip.installed` started failing (on 16.04 at least) with Salt unable to found pip. Manual testing shows pip is installed system-wide and works fine, so I'm replacing these broken states with commands.

PR #187, opened 2018-10-25
Remove leftover pip.installed

giorgiosironi:
Follow-up to https://github.com/elifesciences/builder-base-formula/pull/186 with the bits not strictly necessary for elife-xpub.

### DIFF
--- a/elife/pypi.sls
+++ b/elife/pypi.sls
@@ -15,6 +15,5 @@ pypi-credentials-ubuntu-user:
             - deploy-user
 
 twine:
-    pip.installed:
-        - pkgs:
-            - twine
+    cmd.run:
+        - name: pip install twine
--- a/elife/uwsgi.sls
+++ b/elife/uwsgi.sls
@@ -18,8 +18,8 @@ uwsgi-pkg:
 {% else %}
 # apps should be installing and using their own version of uwsgi
 uwsgi-pkg:
-    pip.installed:
-        - name: uwsgi >= 2.0.8
+    cmd.run:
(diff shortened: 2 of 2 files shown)


elife/pypi.sls
```diff
@@ -15,6 +15,5 @@ pypi-credentials-ubuntu-user:
             - deploy-user
 
 twine:
-    pip.installed:
-        - pkgs:
-            - twine
+    cmd.run:
+        - name: pip install twine
```
diversemix: Did you want to pin a version of this? (ref: https://pypi.org/project/twine/#history )

giorgiosironi: At the moment it doesn't pin a version, by default I'm keeping the same behavior. Perhaps pinning `1.x` would be a good idea.

giorgiosironi:
(18.04 to be fixed later, also thanks to this changes)

elife/uwsgi.sls
```diff
@@ -18,8 +18,8 @@ uwsgi-pkg:
 {% else %}
 # apps should be installing and using their own version of uwsgi
 uwsgi-pkg:
-    pip.installed:
-        - name: uwsgi >= 2.0.8
+    cmd.run:
+        - name: pip install "uwsgi>=2.0.8"
```
lsh-0: we should figure out what projects are still using the global uwsgi and update them.

giorgiosironi: some that have been moved to Docker containers naturally had to use their own `uwsgi` in the virtualenv

PR #189, opened 2018-11-02
Guarantee minimum AWS cli 1.16.*

giorgiosironi:
(contains `aws eks update-kubeconfig`)

### DIFF
--- a/elife/aws-cli.sls
+++ b/elife/aws-cli.sls
@@ -1,4 +1,4 @@
 aws-cli:
     cmd.run:
-        - name: pip install awscli
+        - name: pip install 'awscli~=1.16'
 


elife/aws-cli.sls
```diff
@@ -1,4 +1,4 @@
 aws-cli:
     cmd.run:
-        - name: pip install awscli
+        - name: pip install 'awscli~=1.16'
```
thewilkybarkid: Not sure about pip, but that syntax elsewhere (eg Composer) would mean `>= 1.16 && < 2.0`, not `1.16.*`...

giorgiosironi: Yes, it's [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release). We need 1.16 to get a fully implemented `aws eks`; newer `1.*` are welcome to avoid having to manually update in the future as the cli tool is a very stable project. We do something similar with composer which always self-updates but not to a new major version which would break it (I think).

PR #194, opened 2018-11-22
Extract first nginx macro

giorgiosironi:
Black magic to extract [this other piece of black magic](https://github.com/elifesciences/lax-formula/blob/master/salt/lax/config/etc-nginx-sitesavailable-lax.conf#L18-L25) and make it reusable from other formulas (in this case `journal-cms`).

[Macros official documentation](https://docs.saltstack.com/en/2017.7/topics/jinja/index.html#macros) is available. The backlinked PR on `lax-formula` shows how to use this.

### DIFF
--- a/elife/nginx-macros.sls
+++ b/elife/nginx-macros.sls
@@ -1,0 +1,10 @@
+{% macro consumer_groups_filter(users) -%}
+variables_hash_max_size 2048; # to fit this map
+map_hash_bucket_size 128;
+map $http_authorization $consumer_groups_filtered {
+    default    "";
+    {%- for i, user in users.items() -%}
+    "Basic {{ salt['hashutil.base64_b64encode'](user['username'] ~ ':' ~ user['password']) }}" $http_x_consumer_groups;
+    {%- endfor -%}
+}
+{%- endmacro -%}
--- a/elife/nginx-macros.sls
+++ b/elife/nginx-macros.sls
@@ -1,4 +1,8 @@
 {% macro consumer_groups_filter(users) -%}
+# Authenticates the api-gateway.
+# Copies a filtered version of the api-gateway X-Consumer-Groups header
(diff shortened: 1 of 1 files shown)


lsh-0:
I've never seen any of this before. The stuff in the lax formula has no explanation and I don't remember reviewing it. I have nfi what you are doing.

giorgiosironi:
[It goes back to 2017](https://github.com/elifesciences/lax-formula/pull/12). Longer explanation follows.

This is how an article page gets its (private) data from `lax`:
```
journal -> api-gateway -> lax
```
In particular, the HTTP headers are:
```
journal --Authorization:journalapikey--> api-gateway --X-Consumer-Groups:view-unpublished-articles--> lax
```
However, if `lax` is publicly accessible, someone could pass `X-Consumer-Groups` directly. So instead we add authentication between `api-gateway` and `lax`:
```
journal --Authorization:journalapikey--> api-gateway --X-Consumer-Groups:view-unpublished-articles,Authorization:apigatewayandlaxsharedsecret--> lax
```
This nginx configuration kills `X-Consumer-Groups` if `apigatewayandlaxsharedsecret` is not there, filtering it out.

Lately `journal-cms` has got the same problem: being accessed through `api-gateway` but also being exposed directly. So this extraction will allow it to use the same pattern without copying and pasting. I'll add some comments as this is pretty obscure, I agree.

lsh-0:
ah, there I am giving my tick of approval. I didn't find this yesterday when going through that file's commit history.

That PR has a good explanation:

> This method replaces the need to check the ip address of the caller.
> Nginx strips the X-Consumer-Groups header, letting it through only if the caller satisfied the basic authentication. Only the api-gateway should know this secret.

giorgiosironi:
(related issue is in the Todo column, will get back to this when picking it up)

PR #201, opened 2019-01-21
Feature: support for pulling private images from dockerhub

lsh-0:


lsh-0:
getting this error:
https://github.com/saltstack/salt/issues/49967
```
[3.80.220.222] out: ----------
[3.80.220.222] out:           ID: salt-docker packages
[3.80.220.222] out:     Function: pip.installed
[3.80.220.222] out:         Name: docker-py>=1.4.0
[3.80.220.222] out:       Result: False
[3.80.220.222] out:      Comment: An exception occurred in this state: Traceback (most recent call last):
[3.80.220.222] out:                 File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1879, in call
[3.80.220.222] out:                   **cdata['kwargs'])
[3.80.220.222] out:                 File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1826, in wrapper
[3.80.220.222] out:                   return f(*args, **kwargs)
[3.80.220.222] out:                 File "/usr/lib/python2.7/dist-packages/salt/states/pip_state.py", line 682, in installed
[3.80.220.222] out:                   out = _check_pkg_version_format(pkg)
[3.80.220.222] out:                 File "/usr/lib/python2.7/dist-packages/salt/states/pip_state.py", line 149, in _check_pkg_version_format
[3.80.220.222] out:                   install_req = InstallRequirement.from_line(pkg)
[3.80.220.222] out:               AttributeError: type object 'InstallRequirement' has no attribute 'from_line'
[3.80.220.222] out:      Started: 04:21:55.747623
[3.80.220.222] out:     Duration: 363.791 ms
[3.80.220.222] out:      Changes:   
```

my local version of Ubuntu 16.03 is running Salt `2017.7.5`, I *think* this has been fixed in `2017.7.9`

pillar/dockerhub.sls
```diff
@@ -0,0 +1,6 @@
+docker-registries:
+    # Dockerhub
+    # https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.dockerng.html#authentication
+    hub: 
+        username: foo
```
giorgiosironi: I think this is implemented in https://github.com/elifesciences/builder-base-formula/blob/master/elife/docker-push.sls as authentication should cover both pushing and accessing private images. That file would have to be renamed.

giorgiosironi:
> my local version of Ubuntu 16.04 is running Salt 2017.7.5, I think this has been fixed in 2017.7.9

I started to get wary about installing libraries for the system Python (which I think Salt is using) to execute Salt modules: at least they could keep them in their own virtualenv like other Python applications. Back in the day I went for using the `docker` command line which was already installed.

lsh-0:
I know what you mean about Python and installing things outside of the virtualenv. Tiresome, buggy and fragile. 

It looks like you've already got a mechanism that covers pulling from a registry that requires authentication. I'll close this and update the formula accordingly

PR #203, opened 2019-01-30
Feature: change in certificate triggers an nginx restart

lsh-0:


elife/certificates.sls
```diff
@@ -50,6 +50,10 @@ web-complete-cert:
     cmd.run:
         - name: cat certificate.crt fullchain.pem > certificate.chained.crt && chgrp {{ pillar.elife.certificates.username }} certificate.chained.crt
         - cwd: /etc/certificates/
+        # only trigger state if either of these two files have changed
```
giorgiosironi:
```
If a state should only execute when another state has changes, and otherwise do nothing, the new onchanges requisite should be used instead of watch. watch is designed to add additional behavior when there are changes, but otherwise the state executes normally.
```
This seems to fit what we need here: only re-execute the `cat` in this case.

elife/nginx.sls
```diff
@@ -94,6 +94,11 @@ nginx-server-service:
             # this sounds like a good thing, but some conf files reference external files that may not be 
             # ready when they appear in here. explicit requisites/watches are better.
             #- file: /etc/nginx/sites-enabled/*
+{% if salt['elife.cfg']('cfn.outputs.DomainName') %}
```
giorgiosironi: I discovered this file includes `elife.certificates`: https://github.com/elifesciences/builder-base-formula/blob/5d8102bed07710a7bc3a41d8c57576c789c0c0c7/elife/nginx.sls#L75
If this is correct, then `elife.certificates` should be removed from https://github.com/elifesciences/builder-private/blob/master/salt/top.sls#L33 right?

The `if` becomes the same as https://github.com/elifesciences/builder-base-formula/pull/203/files#diff-35caf6f7a7a31050fce46cdeb1e29a5fL24 so it should be safe for now. However could we do this with a `listen_in` from `web-complete-cert`? This is how [this file does it](https://github.com/elifesciences/builder-base-formula/blob/5d8102bed07710a7bc3a41d8c57576c789c0c0c7/elife/nginx.sls#L23) and it could work without this `if`.

elife/certificates.sls
lsh-0: yes, I read and re-read that passage quite a bit ;)

elife/nginx.sls
lsh-0: that's possible, but certificates.sls has no dependency on any webserver or it's configuration. This was back when we had Apache and Nginx in play. Now that civi and the last app using Apache has been removed the distinction doesn't seem that important any more. 

Can you think of a case where we would need the SSL certificates on a server but not a web server? An FTP or mail server or something?

The master-server project has a dependency on certificates.sls as well, not sure why.

giorgiosironi: We do use them with ELB/CloudFront/Fastly but Salt isn't involved in that. 

The corner cases that come to mind are:
- local usage: web servers are there but no certificates (or fake ones maybe) are used
- [https://github.com/elifesciences/redirects-formula/blob/master/salt/example.top](redirects) which uses its own certificates because it has different domains to cover. This one pushes for decoupling of the two more than for the use of `elife.certificates` without `elife.nginx`.

giorgiosironi: Another case:
- `elife.vault` relies on `elife.certificates` only, https://github.com/elifesciences/master-server-formula/blob/27cf38250da0a339cbb591d400b00249ca0bbdec/salt/master-server/config/etc-vault.hcl
so the direction of the dependency is correct `elife.nginx`, depends on `elife.certificates`

lsh-0: I'll merge this, but we could consider self-signed certificates to get rid of that great big `if` statement and ensure a certificate is always there

PR #207, opened 2019-03-08
Feature: sets 'fs_passno' to '2' on external volumes 

lsh-0:
This is so they get checked (if necessary) on reboot.

I've never done this with AWS/ec2 before and the volumes we have tend to be smallish compared so I'm not sure how much downtime might be incurred while the fs is chk'ed.

### DIFF
--- a/elife/external-volume.sls
+++ b/elife/external-volume.sls
@@ -29,6 +29,7 @@ mount-external-volume:
         - mkmnt: True
         - opts:
             - defaults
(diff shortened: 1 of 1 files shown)


giorgiosironi:
> how much downtime might be incurred while the fs is chk'ed

If it's a reboot, it's already incurring downtime so might as well do the rest of the maintenance. If the project needs to be highly available it will have multiple instances.

giorgiosironi:
Perhaps do a reboot test with a masterless `lax` using this?

PR #208, opened 2019-03-11
Fix: painless systemd+uwsgi restarts

lsh-0:
I don't think anything is using systemd+uwsgi, but if they are then they're probably having problems with `systemctl restart servicename` timing out (minutes?), force killing the process and only then restarting.

This is the approach recommended by uwsgi, albeit not using systemd templates: 
https://uwsgi-docs.readthedocs.io/en/latest/Systemd.html#one-service-per-app-in-systemd

These `.socket` files can be treated just like services by Salt.

The log file appears to be owned and writeable by root only, even though uwsgi is supposed to be dropping privileges to the www-user, so not sure what is going on there. I think deferring the privilege dropping to systemd is the best choice here. uwsgi's 'emperor' mode that acts as a process manager for multiple uwsgi applications may be blurring the line between upstart/systemd and uwsgi. Cool feature though.

I have elife-metrics using [this approach](https://github.com/elifesciences/elife-metrics-formula/pull/3) nicely.

### DIFF
--- a/elife/config/lib-systemd-system-uwsgi-service.service
+++ b/elife/config/lib-systemd-system-uwsgi-service.service
@@ -6,6 +6,8 @@ After=network.target
 WantedBy=multi-user.target
 
 [Service]
+User={{ pillar.elife.webserver.username }}
+Group={{ pillar.elife.webserver.username }}
 KillSignal=SIGQUIT
 Type=notify
 Restart=on-failure
@@ -14,5 +16,5 @@ Environment="LANG=en_US.UTF-8"
 Environment="NEW_RELIC_CONFIG_FILE={{ folder }}/newrelic.ini"
 ExecStart={{ folder }}/venv/bin/newrelic-admin run-program {{ folder }}/venv/bin/uwsgi --enable-threads --ini {{ folder }}/uwsgi.ini
 {% else %}
-ExecStart={{ folder }}/venv/bin/uwsgi --ini {{ folder }}/uwsgi.ini
+ExecStart={{ folder }}/venv/bin/uwsgi --ini {{ folder }}/uwsgi.ini --logto /var/log/uwsgi.log
 {% endif %}
--- a/elife/uwsgi.sls
+++ b/elife/uwsgi.sls
@@ -10,23 +10,37 @@
 # -- may contain a newrelic.ini file
 #
 
-{% if salt['grains.get']('osrelease') != "14.04" %}
-uwsgi-pkg:
-    pkg.installed:
-        - pkgs:
-            - gcc # needed for building uwsgi
-{% else %}
-# apps should be installing and using their own version of uwsgi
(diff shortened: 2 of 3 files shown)


giorgiosironi:
> I don't think anything is using systemd+uwsgi

`profiles` was probably using this at some point, but got shifted to containers.

elife/config/lib-systemd-system-uwsgi-service.service
```diff
@@ -14,5 +16,5 @@ Environment="LANG=en_US.UTF-8"
 Environment="NEW_RELIC_CONFIG_FILE={{ folder }}/newrelic.ini"
 ExecStart={{ folder }}/venv/bin/newrelic-admin run-program {{ folder }}/venv/bin/uwsgi --enable-threads --ini {{ folder }}/uwsgi.ini
 {% else %}
-ExecStart={{ folder }}/venv/bin/uwsgi --ini {{ folder }}/uwsgi.ini
+ExecStart={{ folder }}/venv/bin/uwsgi --ini {{ folder }}/uwsgi.ini --logto /var/log/uwsgi.log
```
giorgiosironi: `logto` should be in `uwsgi.ini` right?

giorgiosironi:
> uwsgi's 'emperor' mode that acts as a process manager

We don't use emperor mode anywhere afaik

elife/config/lib-systemd-system-uwsgi-service.service
lsh-0: it was in uwsgi.ini, I moved it into the service definition. It's something that shouldn't be configured on a per-app basis and the ownership of the log file changes between 14.04 and 16.04 too.

elife/uwsgi.sls
```diff
@@ -69,9 +88,25 @@ uwsgi-service-{{ name }}:
...
+    file.managed:
+        - name: /lib/systemd/system/uwsgi-{{ name }}.socket
+        - source: salt://elife/config/lib-systemd-system-uwsgi.socket
+        - template: jinja
+        - require:
+            - uwsgi-pkg
+            - uwsgi-params
+        - context:
+            name: {{ name }}
+        # necessary?
```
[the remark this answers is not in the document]
lsh-0: I was wondering more if sockets are affected by daemon-reload like services are. I'm thinking they probably are, certainly doesn't hurt to have this

lsh-0:
> We don't use emperor mode anywhere afaik

we don't, my point was that because uwsgi *has* an emperor mode, the nature of the configuration it presents to us for configuration overlaps with other process managers like upstart and systemd. Because of that we need to make decisions about should live with the service/socket definitions and what lives in the .ini file. Stuff like where the stderr/stdout goes should live with the service definition I think, like upstart did. The number of processes, size of thread pool, timeouts etc is probably best left to uwsgi.

PR #211, opened 2019-03-18
Expand retry script with interval

giorgiosironi:
Helps https://github.com/elifesciences/journal-formula/pull/89


elife/utils/retry
```diff
@@ -2,18 +2,20 @@
 set -e
 
 if [ "$#" -lt 1 ]; then
-    echo "Usage: retry SCRIPT_PATH [MAXIMUM_TIMES]"
-    echo "Example: retry ./script.sh 3"
+    echo "Usage: retry SCRIPT_PATH [MAXIMUM_TIMES] [INTERVAL]"
+    echo "Example: retry ./script.sh 3 30"
```
thewilkybarkid: Seconds?

giorgiosironi: Yes, since they are passed to `sleep`

PR #216, opened 2019-03-22
Feature: stops and disables snapd service in 18.04 instances

lsh-0:


giorgiosironi:
We don't use `snap` on servers, and it shouldn't be available anyway for installation of software outside what the formula specifies.

elife/base.sls
```diff
@@ -69,3 +69,10 @@ ubuntu-user:
         - shell: /bin/bash
         - groups:
             - sudo
+
+# unnecessary always-on new container service in 18.04 that nothing uses
```
giorgiosironi: wrap in an `if` on the OS version for extra safety?

PR #219, opened 2019-04-01
Feature: application certificate overrides

lsh-0:
* adds support for applications to switch in their own certificates by overriding a pillar default
* default set of certificates are those provided by builder-private ('elife')

### DIFF
--- a/elife/certificates.sls
+++ b/elife/certificates.sls
@@ -22,18 +22,20 @@ web-certificates-dir:
             - web-certificates-group
 
 {% if salt['elife.cfg']('cfn.outputs.DomainName') %}
+{% set app = pillar.elife.certificates.app or 'elife' %}
+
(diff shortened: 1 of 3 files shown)


elife/certificates.sls
```diff
@@ -22,26 +22,28 @@ web-certificates-dir:
             - web-certificates-group
 
 {% if salt['elife.cfg']('cfn.outputs.DomainName') %}
+{% set app = pillar.elife.certificates.app or 'elife' %}
```
giorgiosironi: do we actually need the default `or 'elife'`? `pillar/elife.sls` is always going to provide the `elife` value

lsh-0: you're right, it is redundant. I felt better safe than sorry should there be a gap between builder-base being merged in before the builder-private change

PR #220, opened 2019-04-02
Feature: adds missing description of newrelic_python pillar data

lsh-0:
if we want article-scheduler to have newrelic support (we may not) then I'll need to make newrelic_python.sls multi-app like the `uwsgi` pillar data

### DIFF
--- a/pillar/elife.sls
+++ b/pillar/elife.sls
@@ -143,6 +143,11 @@ elife:
         rest_api_key: null
         insights_api_key: null
 
+    newrelic_python: {}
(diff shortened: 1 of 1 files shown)


giorgiosironi:
> if we want article-scheduler to have newrelic support

gosh no, too legacy to intervene on it in any way than security fixes

lsh-0:
ha - the impression of you saying 'gosh' is just too funny ... you've been hanging around fusty english people too long :D

giorgiosironi:
I also say 'geez' but I'm not sure it balances it out

PR #223, opened 2019-04-05
Fix: newrelic_python further restricted

lsh-0:
* `newrelic_python.sls` states only available if newrelic enabled AND newrelic_python configuration available.

we had a highstate failure in `elife-dashboard--prod` with:

```
local:
    Data failed to compile:
----------
    Rendering SLS 'base:elife.newrelic-python' failed: Jinja variable 'dict object' has no attribute 'newrelic_python'
```

because the `elife.newrelic_python` state had been included in the `top.sls` file but no `pillar.elife.newrelic_python` configuration made available.

I know it says *right at the top of the file* that the state file cannot be included without also specifying [...] but this would change it from a hard failure with an esoteric message to a failed state with a descriptive error message.

Not sure which approach is better, feel free to close if you prefer the original way

### DIFF
--- a/elife/newrelic-python.sls
+++ b/elife/newrelic-python.sls
@@ -9,7 +9,7 @@
 # - pillar.elife.newrelic_python.service 
 #     the name of a service.running state that should be restarted
 
-{% if pillar.elife.newrelic.enabled %}
+{% if pillar.elife.newrelic.enabled and pillar.elife.newrelic_python %}
 newrelic-python-license-configuration:
     cmd.run:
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Retrying build to see if we can keep it green thanks to https://github.com/elifesciences/heavybox-formula/pull/10

giorgiosironi:
The answer is: everything breaks, of course.

giorgiosironi:
Back to :green_apple:

lsh-0:
huh. had forgotten all about this

PR #224, opened 2019-04-05
Remove race conditions from docker-prune

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4850

This version is still aggressive on image removal as it will remove any image as long as there is no container running with it. The date filter is not really useful because it refers to the image _creation_ date, which for any deployment may be far in the past (e.g. `elifesciences/sciencebeam:0.0.1`)

We counteract this fact by only cleaning if it is necessary i.e. if more than 80% of space or inodes are being used at the moment.

### DIFF
--- a/elife/docker-scripts/docker-prune
+++ b/elife/docker-scripts/docker-prune
@@ -3,6 +3,20 @@ set -e
 
 # default: 14 days
 hours_limit="${1:-336}"
+# default: 80%
+usage_limit="${2:-80}"
+
+if [ -z "$FORCE" ]; then
+    echo "Usage limit: ${usage_limit}%"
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Typical runs:
```
$ docker-prune 336 80
Usage limit: 80%
Disk space used:   7%
Disk inodes used:    4%
Not necessary to run because of low usage
$ docker-prune 336 5
Usage limit: 5%
Disk space used:   7%
Disk inodes used:    4%
Clean up all stopped containers older than the last 336 hours
Total reclaimed space: 0B
Clean up all unused networks older than the last 336 hours
Clean up all unused volumes
Total reclaimed space: 0B
Clean up all images not referenced by a container
Total reclaimed space: 0B
```

PR #226, opened 2019-04-09
re-enables postgresql pg_hba conf on mixed local+rds instances

lsh-0:
no harm in that conf file sitting there if the service is dead

this is for elife-dashboard where, for some unholy reason, this has gone undetected since *forever*. 

cc @giorgiosironi - if this is green I'll merge it in so I can finish off `--end2end` and (hopefully) `--prod` before the UK wakes up for work

### DIFF
--- a/elife/postgresql.sls
+++ b/elife/postgresql.sls
@@ -48,6 +48,7 @@ postgresql:
             - pgpass-file
 {% endif %}
 
+# TODO: remove this, not used in 16.04+, uncertain about 14.04
 {% if not salt['elife.cfg']('cfn.outputs.RDSHost') %}
 postgresql-init:
     file.managed:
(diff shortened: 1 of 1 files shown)


elife/postgresql.sls
```diff
@@ -48,6 +48,7 @@ postgresql:
             - pgpass-file
 {% endif %}
 
+# TODO: remove this, not used in 16.04+, uncertain about 14.04
```
lsh-0: 18.04 has a systemd service file that is used

lsh-0:
I'm not so sure this is the answer anymore

lsh-0:
while I think this is probably a good set of changes for postgresql, I'd rather not introduce any instability right this minute. Closing.

PR #227, opened 2019-04-09
postgresql 'pg_hba.conf' file always present, even if service is dead

lsh-0:
on dual local+rds instances, the local `pg_hba.conf` file was left unmanaged. 

this file had probably been tweaked by hand (changing 'peer' to 'md5') to get it 'working' previously


giorgiosironi:
Since this is related to the client rather than the server, it is correct for it to be managed in all cases.

PR #228, opened 2019-04-18
uwsgi, fixes logging and logrotation introduced with the 1604/1804 update

lsh-0:
log monitoring and logrotate rules for each uwsgi service.

this is the cause of what is currently affecting `elife-metrics--prod`


lsh-0:
hm. this isn't working on elife-metrics. 'appname' is apparently undefined. investigating ...

PR #230, opened 2019-04-26
Fix: adds 'openjdk8.sls' statefile as a replacement for 'java8.sls'

lsh-0:
oracle java 8 on 14.04 [is now completely broken](https://launchpad.net/~webupd8team/+archive/ubuntu/java), rather than just 'kinda broken', and is preventing the `search` project from passing tests.

This new state file is intended as a replacement for the `java8.sls`.

Works for me locally on 14.04 and 18.04 using a tweaked basebox.

### DIFF
--- a/elife/openjdk8.sls
+++ b/elife/openjdk8.sls
@@ -19,6 +19,8 @@ oracle-java8-installer-removal:
 openjdk8 ppa:
     pkgrepo.managed:
         - ppa: jonathonf/openjdk
+        - require_in:
+            - pkg: java8
         
 {% endif %}
 
--- a/elife/openjdk8.sls
+++ b/elife/openjdk8.sls
@@ -1,0 +1,35 @@
+# 14.04, 16.04 and 18.04 have a "openjdk-8-jre-headless" package
+# 14.04 requires a ppa
+
+# remove oracle java (if it exists)
+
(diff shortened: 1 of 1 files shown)


elife/openjdk8.sls
```diff
@@ -0,0 +1,37 @@
...
+    pkg.purged:
+        - name: oracle-java8-installer
+        - require:
+            - oracle-ppa-removal
+
+{% if salt['grains.get']('osrelease') == "14.04" %}
+
+openjdk8 ppa:
+    pkgrepo.managed:
+        - ppa: jonathonf/openjdk
```
giorgiosironi: does this require a key? e.g. https://github.com/elifesciences/builder-base-formula/blob/01b601ead7d331abb4d0714e5d597bda38a123ed/elife/gcloud.sls#L4-L16

lsh-0: [doesn't appear to:](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkgrepo.html#salt.states.pkgrepo.managed)

> APT-BASED SYSTEMS
> ppa
>On Ubuntu, you can take advantage of Personal Package Archives on Launchpad simply by specifying the user and archive name. The keyid will be queried from launchpad and everything else is set automatically.

PR #231, opened 2019-04-30
Feature: removes the amazon-ssm-agent 'snap'

lsh-0:
and purges snapd, the snap daemon

### DIFF
--- a/elife/base.sls
+++ b/elife/base.sls
@@ -40,6 +40,8 @@ base-purging:
         - pkgs:
             - puppet
             - snapd
+        - require:
+            - base
 
 base-vim-config:
     file.managed:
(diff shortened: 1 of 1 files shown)


lsh-0:
apparently if you don't remove the 'snap' files and cache dir, then apt can't purge properly

lsh-0:
looks like reverse requisites aren't being obeyed. again. fml

elife/base.sls
```diff
@@ -71,10 +74,28 @@ ubuntu-user:
...
             - hash snap
+        - require:
+            - amazon-ssm-agent-snap-removal
+
+    cmd.run:
+        - name: rm -rf /var/cache/snapd
+        - require:
+            - service: snapd
+        - require_in:
+            - base-purging
```
giorgiosironi: lore says that you need to specify both the state type and its name in reverse requisites, `cmd: base-purging` (or whatever it is)

lsh-0: ah - I'll give that a shot, cheers.

lsh-0: whaddya know, it's green

PR #234, opened 2019-05-14
Feature: systemd multi-process test demonstration

lsh-0:
cc @giorgiosironi 

a demonstration of managing a group of services using a `.target` unit. It works really well but is not perfect yet.

I want to:
* avoid restarting the service on daily highstate
* grow/shrink the number of processes and have the right things enabled/disabled/restarted

Turns out salt mangles service names that look like "service@{x..y}" so can't use the `service` state with the individual processes, which shouldn't be a problem if they're being treated as a group using the controller service.

### DIFF
--- a/elife/multi-process-test.sls
+++ b/elife/multi-process-test.sls
@@ -1,0 +1,48 @@
+
+# test that N processes are brought using a single service state
+# each process simply writes it's identifier and the time to the syslog
+
+{% set process = "test" %}
+{% set num_processes = 10 %}
+
+/opt/{{ process }}.py:
+    file.managed:
+        - source: salt://elife/scripts/test.py
+
+{{ process }}:
+    file.managed:
+        - name: /lib/systemd/system/{{ process }}@.service
+        - source: salt://elife/config/lib-systemd-system-test.service
+        - template: jinja
+        - context:
+            process: {{ process }}
+        - require:
+            - /opt/{{ process }}.py
+
+    # neccessary because salt mangles "service@{x..y}" to "service@\x7bx..y\x7d.service" 
+    # systemd will see that as just a single instantiated service
+    cmd.run:
+        - name: |
+            set -e
+            systemctl disable {{ process }}@{0..99} # disable all instances (up to 99). implicit reload
+            systemctl enable {{ process }}@{0..1} # enable just the range we're after. implicit reload
+        - require:
+            - file: {{ process }}
+        # only run command if there are disabled instances in our range
+        # todo: only runs if range grows. if range shrinks to 0..0 from 0..1 then this command won't be triggered
+        - onlyif:
+            - systemctl is-enabled {{ process }}@{0..1} | grep -q 'disabled'
+
+{{ process }}-controller.target:
+    file.managed:
+        - name: /lib/systemd/system/{{ process }}-controller.target
+        - source: salt://elife/templates/process-controller.target
+
+    service.running:
+        - enable: true
+        - require:
+            - file: {{ process }}-controller.target
+            - cmd: {{ process }}
+        # restarts target (and all processes that need that target) when processes enabled
+        - watch:
+            - cmd: {{ process }}
--- a/elife/multi-process-test.sls
+++ b/elife/multi-process-test.sls
@@ -1,15 +1,18 @@
 
-# test that N processes are brought using a single service state
-# each process simply writes it's identifier and the time to the syslog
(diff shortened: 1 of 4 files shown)


lsh-0:
I just put an entry in my basebox-formula's `top` file and brought it up with vagrant.

useful commands:
* `systemctl` to see which units have been loaded
* `systemctl enable service@{0..1}` to enable two processes
* `systemctl disable service@{0..99}` to disable all enabled processes
* `journalctl -f ` to watch the processes write to syslog

elife/multi-process-test.sls
```diff
@@ -0,0 +1,48 @@
...
+            process: {{ process }}
+        - require:
+            - /opt/{{ process }}.py
+
+    # neccessary because salt mangles "service@{x..y}" to "service@\x7bx..y\x7d.service" 
+    # systemd will see that as just a single instantiated service
+    cmd.run:
+        - name: |
+            set -e
+            systemctl disable {{ process }}@{0..99} # disable all instances (up to 99). implicit reload
```
giorgiosironi: could it just `stop` and `start` rather than `disable` and `enable`? The difference is whether it affects what happens at boot and so the time interval in which `cmd.run` is executed it doesn't make a difference.

giorgiosironi: If there are many processes and `disable` waits for them, there may be a performance problem. In the Upstart script we send stopping signals for all of them asynchronously (no waiting) and then wait for termination of everything as a second step.

giorgiosironi: Maybe this doesn't actually stop and start them now, which would make sense.

elife/multi-process-test.sls
```diff
@@ -0,0 +1,48 @@
...
+            systemctl enable {{ process }}@{0..1} # enable just the range we're after. implicit reload
```
giorgiosironi: `1` should be `num_processes`?
Same for the `onlyif`.

elife/multi-process-test.sls
```diff
@@ -0,0 +1,48 @@
...
+        # todo: only runs if range grows. if range shrinks to 0..0 from 0..1 then this command won't be triggered
+        - onlyif:
+            - systemctl is-enabled {{ process }}@{0..1} | grep -q 'disabled'
+
+{{ process }}-controller.target:
+    file.managed:
+        - name: /lib/systemd/system/{{ process }}-controller.target
+        - source: salt://elife/templates/process-controller.target
+
+    service.running:
```
giorgiosironi: This state does not restart the processes:
```
          ID: test-controller.target
    Function: service.running
      Result: True
     Comment: The service test-controller.target is already running
     Started: 10:21:43.602231
    Duration: 30.64 ms
     Changes:  
```
but
```
sudo systemctl restart test-controller.target
```
does. Include that with a `cmd.run`?

lsh-0: it's not going to restart the state if it's already running, no.

you would want to restart the state if the number of processes running is different from what you intended, whether one has died or more added/removed.

this target/unit/state/service thing doesn't care if the processes that depend on it are there or not, so it's probably not ever going to fail in coming up or being restarted.

elife/multi-process-test.sls
lsh-0: this example isn't quite finished yet. enable and disable for template services determines how many are started when the controller becomes available. 

if we enable 5 processes and restart the controller, then 5 processes will be started.

if we change that to 3 processes and restart the controller, we want just 3 processes running but we still have 2 others that weren't disabled.

you can target processes individually with `service@n`, and we do want some feedback in the formula if the service failed to start, so we could have a `for` loop of `service.running` states in the formula.

lsh-0: > If there are many processes and disable waits for them

it doesn't. `enable` and `disable` are a bit like `a2enmod` and `a2dismod` - they're just wrappers around creating symlinks.

when the controller state changes these processes are brought up or stopped asynchronously

lsh-0:
ok, I've finished the example now and incorporated the feedback.

* there are `service.running` states for each of the N processes we want running
* there is a state that brings the enabled and running processes inline with the intended number.

I've separated out the test script and it's service file from the rest of the file - I'm going to make it behave like the uwsgi service. It's gotten too complex to just copy and paste between formulas

lsh-0:
closing in favour of: https://github.com/elifesciences/builder-base-formula/pull/235

PR #235, opened 2019-05-15
Feature: multiservice

lsh-0:
introduces `multiservice.sls` that treats multiple services as a group and provides a target for salt to stop/start/restart all running instances of this service.

pillar configuration is similar to `uwsgi`

formulas that implement this will need to provide their own service template file and the number of processes they want running.


giorgiosironi:
> formulas that implement this will need to provide their own service template file and the number of processes they want running.

That's expected, so ok.

pillar/elife.sls
```diff
@@ -233,6 +233,15 @@ elife:
             #profiles:  
             #    folder: /srv/profiles
 
+    multiservice:
+        services: {}
+            #whatever:
```
giorgiosironi: `myservice`?

pillar/elife.sls
```diff
@@ -233,6 +233,15 @@ elife:
             #profiles:  
             #    folder: /srv/profiles
 
+    multiservice:
```
giorgiosironi: should this top key include `systemd` as a keyword, since this solution is specific to it? Or is it intended as being hidden from the user?

elife/multiservice.sls
```diff
@@ -0,0 +1,55 @@
...
+            - service: {{ process }}-controller.target
+
+# ensure N processes are running 
+{% for i in range(0, num_processes) %}
+{{ process }}@{{ i }}:
+    service.running:
+        - enable: true # redundant, but can't hurt
+        - require:
+            - {{ process }}-set-enabled
+            - {{ process }}-controller.target
```
giorgiosironi: trying this out locally in combination with https://github.com/elifesciences/journal-cms-formula/pull/54 and the 18.04 Vagrant box. Main concern is that during a new build all processes are restarted, I don't see states or `systemctl` commands that accomplish that :thinking:

giorgiosironi: Can't get it to work quickly, better to test it with the `basebox` one.

giorgiosironi: Confirmed: processes are not restarted on a new highstate.

elife/multiservice.sls
```diff
@@ -0,0 +1,55 @@
...
+{% for process, opts in pillar.elife.multiservice.services.items() %}
+    {% set num_processes = opts["num_processes"] %}
+
+# ensure the target state is available and running
+{{ process }}-controller.target:
+    file.managed:
+        - name: /lib/systemd/system/{{ process }}-controller.target
+        - source: salt://elife/templates/process-controller.target
+
+    service.running:
```
giorgiosironi: this seems too early as at this point the right number of processes hasn't been enabled yet

giorgiosironi: `systemctl restart tester-controller.target` works to restart all processes in parallel (the holy grail), will add it.

elife/multiservice.sls
```diff
@@ -0,0 +1,45 @@
...
+        - source: salt://elife/templates/process-controller.target
+
+# ensure precisely N processes are enabled. restart controller if N has changed
+# enabled processes are started when the controller state changes
+{{ process }}-set-enabled:
+    cmd.run:
+        - name: |
+            set -e
+            systemctl stop {{ process }}@{0..99} # stop any running instances of this service.
+            systemctl disable {{ process }}@ # disable *everything*. implicit reload
```
giorgiosironi: service definition file isn't reloaded if the number of processes doesn't change, while it should happen for any change to that file

elife/multiservice.sls
giorgiosironi: Solved by using an explicit restart of the target.

elife/multiservice.sls
giorgiosironi: Solved by separating the `enable` part so that it is always executed (it's idempotent).

pillar/elife.sls
lsh-0: I don't mind, my imagination was failing me for a while. I think 'multiservice' is short and sweet but it's totally made up. Systemd terminology would 'systemd unit group target' or something equally terse.

elife/multiservice.sls
lsh-0: > Confirmed: processes are not restarted on a new highstate

they're not supposed to! 

these are regular processes that shouldn't be restarted if it's not necessary. if a project wants it's  set of processes restarted once a day on highstate then it can go above and beyond and issue a `systemctl restart myservice-controller` command.

elife/multiservice.sls
lsh-0: I'll take a look at your work today

elife/multiservice.sls
lsh-0: I really think we're on two different wavelengths here. I'm not going to merge the code with that restart statement in there until I understand and agree with what you're trying to do.

giorgiosironi: It's not restarting on a daily highstate that is interesting, but during a deployment.

Existing projects that contain long-running processes rely on the Upstart script [etc-init-multiple-processes-parallel.conf](https://github.com/search?q=org%3Aelifesciences+%22etc-init-multiple-processes-parallel%22&type=Code) to restart all of these processes on each deployment or daily highstate. The one that don't [have workarounds to still restart them all the time](https://github.com/elifesciences/lax-formula/blob/master/salt/lax/bot-lax-adaptor.sls#L175-L177).

It is possible in theory to write enough `watch` statements to trigger restarts only when needed, but in practice it is too complex for projects to implement, so they resort to always restart. The recent case was rotating credentials, with services remaining on the old credentials for a while?

giorgiosironi: Long story short: I am porting the current behavior of the Upstart implementation.

lsh-0: ok, so these processes *are* being restarted on highstate, but only as a side effect because we're not being strict enough in writing watches for things like credentials and configuration changing?

I think this is a lame solution for broken formulas. If all it needs is a restart then there was no need to split the `{{ process }}-set` state into two or remove the `service.running` states. In fact, if we settle on a restart on every highstate and don't care that commands are being unnecessarily run then this whole file could be gutted and replaced with something simpler.

lsh-0: take a look at this PR: https://github.com/elifesciences/builder-base-formula/pull/239

same effect, much simpler, better checks against running processes

PR #236, opened 2019-05-15
Add git author to Jenkins nodes

giorgiosironi:
Since they can commit and push in pipelines, they might as well use a
standard author and email.

This affects `containers` nodes as they are managed by Jenkins and use the `ubuntu` user. `elife-libraries` is already covered by the `elife` user.


giorgiosironi:
PHP is broken on 14.04. Wondering whether to remove it and/or move it to the deprecated build, which doesn't even appear here because it's from `heavybox-formula`.

PR #237, opened 2019-05-16
Make openssl dhparam generation faster

giorgiosironi:
See advice at https://security.stackexchange.com/questions/95178/diffie-hellman-parameters-still-calculating-after-24-hours

Security implications, if any, to be discussed.


giorgiosironi:
(expected 14.04 failure, but haven't checked it in depth)

PR #238, opened 2019-05-17
Feature: gearman, disables installation of upstart script in non-14.04 instances

lsh-0:
it's not dong any harm right now, I'm just trying to clean up the gearman configuration in `search`

### DIFF
--- a/elife/gearman-server.sls
+++ b/elife/gearman-server.sls
@@ -4,6 +4,7 @@ gearman-daemon:
             - gearman-job-server
             - gearman-tools
 
+{% if salt['grains.get']('osrelease') == '14.04' %}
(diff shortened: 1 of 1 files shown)


giorgiosironi:
What do we do with PHP in 14.04? Even when we will have ported away all projects, the state will continue to fail there and breaking the full build for these pull requests.

lsh-0:
I was thinking about that today actually. I didn't have any nice solutions, unfortunately. I figured we put up with it until all projects are ported away and then purge the 14.04 bits. And the php ppa bits.

lsh-0:
or we shift the default AMI to 16.04 for everything?

giorgiosironi:
On the 14.04 PHP status:
https://github.com/elifesciences/heavybox-formula/pull/10

PR #239, opened 2019-05-20
multiservice, simplified states after accepting a restart-on-highstate

lsh-0:
* it takes about 2 seconds to stop, disable, enable and start 10 of those test services.
* it's a bit shitty, but it will always happen, no conditions and no other states where another bit of the logic resides.
* re-introduces `service.running` state. these were intended to catch process failures that systemd doesn't. systemd just starts the process and if it dies immediately, it still reports it as having successfully started. I've added an optional delay that I think one of the processes in journal-cms would benefit from.

if you're happy with my changes, feel free to merge it into `feat-systemd-targets` and then merge that into `master`

### DIFF
--- a/elife/multiservice.sls
+++ b/elife/multiservice.sls
@@ -27,6 +27,7 @@
             {% endif %}
             systemctl start {{ process }}-controller.target
         - require:
+            - {{ process }}-controller.target
             - file: {{ opts["service_template"] }} # name of state that manages the systemd service template file
 
 # 3. broken processes fail highstate as they ordinarily would
--- a/elife/multiservice.sls
+++ b/elife/multiservice.sls
@@ -24,8 +24,8 @@
             systemctl disable {{ process }}@ # disable *everything*. implicit reload
             {% if num_processes > 0 %}
             systemctl enable {{ process }}@{0..{{ num_processes - 1}}} # enable just the range we're after. implicit reload
-            {% endif %}
             systemctl start {{ process }}-controller.target
+            {% endif %}
         - require:
             - {{ process }}-controller.target
             - file: {{ opts["service_template"] }} # name of state that manages the systemd service template file
--- a/elife/multiservice.sls
+++ b/elife/multiservice.sls
@@ -8,45 +8,36 @@
 {% for process, opts in pillar.elife.multiservice.services.items() %}
     {% set num_processes = opts["num_processes"] %}
 
-# ensure the target state is available and running
+# ensure the target state controller is available and running
 {{ process }}-controller.target:
     file.managed:
         - name: /lib/systemd/system/{{ process }}-controller.target
         - source: salt://elife/templates/process-controller.target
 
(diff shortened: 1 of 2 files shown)


elife/multiservice.sls
```diff
@@ -8,45 +8,37 @@
...
-              test {{ num_processes }} != $(systemctl show {{ process }}@{0..99} | grep '^MainPID=[^0]' | wc -l)
-
-{{ process }}-set-enabled:
-    cmd.run:
-        - name: |
-            set -e
             {% if num_processes > 0 %}
             systemctl enable {{ process }}@{0..{{ num_processes - 1}}} # enable just the range we're after. implicit reload
             {% endif %}
+            systemctl start {{ process }}-controller.target
```
giorgiosironi: do we need to put this into the `if` on `num_processes`? Nothing to start in the `0` case, but it will probably do nothing anyway

giorgiosironi: do we need to put this into the `if` on `num_processes`? Nothing to start in the `0` case, but it will probably do nothing anyway

lsh-0: good point, I'll take a second look

lsh-0:
> disabling/enabling does not have overhead I assume

minimal, it's just creating symlinks and doing an implicit daemon-reload I think. The greater overhead lies in stopping the service which I think systemd *does* wait for (unlike starting a service).

elife/multiservice.sls
lsh-0: this looks like a duplicate of the above but could just as well refer to the jinja `for` loop just beneath it. a `range(0, 0)` yields an empty list which is skipped by the template.

other case fixed up though

lsh-0:
alright, when this is green I'll merge it down the line and ready the blocked projects for updates Monday

elife/multiservice.sls
```diff
@@ -8,45 +8,37 @@
...
             systemctl disable {{ process }}@ # disable *everything*. implicit reload
-        - require:
-            - file: {{ opts["service_template"] }} # name of state that manages the systemd service template file
-        - onlyif:
-            # only run command when the expected number of processes (enabled and running) is not equal to the intended 
-            # number of processes. 
-            # use 'is-enabled' to find enabled units. it conveniently returns 'disabled' if all are disabled, so grep for 'enabled'
-            # use 'show' and grep for the pid for running units. it conveniently returns MainPID=0 for enabled-but-not-running units
-            - test {{ num_processes }} != $(systemctl is-enabled {{ process }}@{0..99} | grep 'enabled' | wc -l) || \
-              test {{ num_processes }} != $(systemctl show {{ process }}@{0..99} | grep '^MainPID=[^0]' | wc -l)
```
lsh-0: honestly glad to be rid of these two checks. systemd just kept throwing edge cases at me

lsh-0:
php errors, merging

PR #240, opened 2019-05-20
docker, fixes a problem with installing and configuring the ppa

lsh-0:
I was getting GPG errors despite the GPG import command succeeding


lsh-0:
I just checked the errors, the problem is php as expected and then all the cascading issues

PR #241, opened 2019-05-24
Feature: syslog-ng, fixes typo in uwsgi syslog-ng config file with repeated stmt

lsh-0:


lsh-0:
just php errors. merging

PR #242, opened 2019-06-03
Enable multiservice targets

giorgiosironi:


elife/multiservice.sls
```diff
@@ -24,12 +24,20 @@
...
             {% endif %}
         - require:
             - {{ process }}-controller.target
             - file: {{ opts["service_template"] }} # name of state that manages the systemd service template file
 
+    {% if num_processes > 0 %}
+    service.running:
+        - name: {{ process }}-controller.target
+        - enable: True
+        # results in a recursive requisite
```
lsh-0: weird, I can't see that at a glance (but I completely believe you).

PR #245, opened 2019-06-18
Add s3 charts plugin to Helm

giorgiosironi:
/cc @de-code related only to `elife-xpub`, but may be useful for other projects.

Should be re-tested after https://github.com/elifesciences/heavybox-formula/pull/11

### DIFF
--- a/elife/helm.sls
+++ b/elife/helm.sls
@@ -18,6 +18,10 @@ helm-init:
         - require:
             - helm
 
+make-dependency:
+    pkg.installed:
+        - name: make
+
 helm-s3-plugin:
     cmd.run:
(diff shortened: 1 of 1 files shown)


de-code:
Just to make sure I understand it, this is to be able to use s3 as a charts repo?

And we would prefer that over GitHub because we think it's more reliable? (Or maybe we don't have a preference?)

giorgiosironi:
> Just to make sure I understand it, this is to be able to use s3 as a charts repo?

Yes

> And we would prefer that over GitHub because we think it's more reliable? (Or maybe we don't have a preference?)

It's the same as pulling a Docker image over building it wherever it is needed, the work to pull in dependencies (like a `postgresql` chart) is done only once and the final artifact (`mychart-*.tgz`) is reused.

PR #246, opened 2019-06-19
Feature: psql, psql-appdb, made implicit state dependencies explicit

lsh-0:
bioprotocol-formula encountered case where postgresql-appdb.sls states were being run before anything in postgresql.sls

### DIFF
--- a/elife/postgresql-appdb.sls
+++ b/elife/postgresql-appdb.sls
@@ -33,6 +33,8 @@ db-perms-to-rds_superuser:
             db_name: {{ db_name }}
             app_user_name: {{ app_user_name }}
             app_user_pass: {{ app_user_pass }}
+        - require:
+            - postgresql-ready # builder-base-formula.postgresql
 
 psql-app-db:
     postgres_database.present:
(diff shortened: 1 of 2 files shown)


elife/postgresql-appdb.sls
```diff
@@ -33,6 +33,8 @@ db-perms-to-rds_superuser:
             db_name: {{ db_name }}
             app_user_name: {{ app_user_name }}
             app_user_pass: {{ app_user_pass }}
+        - require:
+            - postgresql-ready # builder-base-formula.postgresql
```
giorgiosironi: isn't that `elife.postgresql` in terms of Salt state files names?

lsh-0: sure. it started off life as "builder-base-formula/elife/postgresql.sls" but I shortened it and figured the intent would be obvious. I'll change it

lsh-0: fixed

giorgiosironi:
(approved anyway)

lsh-0:
(I would have merged anyway ;)

PR #247, opened 2019-06-20
Add Helm S3 plugin to publish charts

giorgiosironi:
Jenkins needs to initialize it and run it as the `jenkins` user rather than messing with files owned by `jenkins` and commands run as `elife`.


giorgiosironi:
Will need to do the same for `kubectl` and `elife-xpub` builds.

PR #248, opened 2019-06-24
Remove support for unsafe TLS 1.0, 1.1

giorgiosironi:
The linked https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html describes 1.2 as the only safe choice.

Impact on browsers is primarily for [IE < 11](https://caniuse.com/#feat=tls1-2) but could be more.

### DIFF
--- a/elife/config/etc-nginx-nginx.conf
+++ b/elife/config/etc-nginx-nginx.conf
@@ -97,7 +97,7 @@ http {
     
     {% if salt['elife.cfg']('cfn.outputs.DomainName') %}
     
-    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+    ssl_protocols TLSv1.2;
 
     ssl_session_cache shared:SSL:10m;
     ssl_session_timeout 10m;


thewilkybarkid:
Presume Fastly (+ELB?) connect with modern support too.

lsh-0:
yes, I've read this article before while setting up our current nginx configuration and decided *not* to go all in.

> describes 1.2 as the only safe choice.

yes, but you do so at the exclusion of older browsers. At the time our SSL was implemented we inadvertently cut off a few percent of IE users and only realised it after the fact in conversation with @davidcmoulton 

I'd like to see some numbers around how much of our current traffic will be excluded with this security bump.

giorgiosironi:
> Presume Fastly (+ELB?) connect with modern support too.

It's a good point that this is irrelevant for `journal`, `iiif` and `api-gateway` because Fastly is what users see, and it only offers TLS 1.2:

> By June 30, 2018, all customers, including those on dedicated endpoints, must have converted to TLS-1.2. Due to the PCI Security Standards Council mandate, older TLS implementations will no longer be supported on Fastly infrastructure on shared or dedicated endpoints. There will be no exceptions made after this date. -- https://www.fastly.com/blog/phase-two-our-tls-10-and-11-deprecation-plan

As for what they connect with, you can constrain them to use a minimum or maximum TLS version but we have not configured that, so they will adapt.

ELBs also can constraint clients, but from a quick view we don't specify anything stricter than [TLS 1.0/1.1/1.2 there](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html).

In the end, this is more relevant for `elife-xpub` because we have [network load balancers (level 4)](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L1942-L1944) there and no CDN, HTTPS is terminated on the single servers. This is due to WebSockets usage IIRC.

Also relevant for projects that expose HTTPS directly from a server, so:
- `journal-cms`
- `pattern-library`
- `elife-dashboard`
- `elife-api` (still alive)
- `elife-alfred`
- `api-dummy`
- `personalised-covers`
- `observer`
- `peerscout`
- `sciencebeam-texture`
- `schematron-validator`
- maybe `profiles` but it should go through `api-gateway` anyway

lsh-0:
> Also relevant for projects that expose HTTPS directly from a server, so:

the external non-elife traffic for all those projects is very minimal. If anything breaks, which I doubt, it's because of an old OS or old code that is under our control.

> elife-api (still alive)

Is joined at the hip to Lens. Since we have no Lens maintainers, we have no way disentangling the two. I'd love to ditch it. Even Arges (not on your list, has a HTTPS server) is fighting extinction right now.

gnott:
An aside,

> elife-api ... [i]s joined at the hip to Lens

If I remember, then only thing the API provides is a Lens readable way to load Glencoe video resources. Changing Lens to read Glencoe directly didn't seem like a priority, unless getting rid of the old API data is more important that doing Lens work. I am only doing bug fixes on Lens, and I've not gone back in to do anything major. I may have some time though in the future.

lsh-0:
it sounds like we have a volunteer :)

I'll find the ticket or create one and move that conversation there.

lsh-0:
(I'm curious, @gnott , how did you get alerted to this ticket? )

gnott:
Most Github issues I have filtering into folders that I do not read, but I read ones related to builder, albeit briefly.

giorgiosironi:
Today is as good a day as any to merge this.

PR #249, opened 2019-07-01
Feature: logrotate, excluding false positives with python

lsh-0:


elife/logrotate.sls
```diff
@@ -1,7 +1,14 @@
+logrotate noise filterer:
+    file.managed:
+        - name: /usr/local/bin/logrotate_noise_filter.py
+        - source: salt://elife/scripts/logrotate_noise_filter.py
```
giorgiosironi: I see `scripts/` doesn't follow the convention that would name this `usr-local-bin-logrotate_noise_filter.py`, no problem.

elife/config/etc-cron.daily-logrotate
```diff
@@ -20,4 +20,8 @@ capture_file=/var/log/daily-logrotate.log
 # logrotate doesn't appear to exit with a non-zero status when there is a 
 # problem rotating a logfile so we have to grep the output.
 # failing to find a problem (|| true) is ok as well
-grep --quiet -i "error:" "$capture_file" && logger "logrotate encountered an error on $(hostname): $capture_file" || true
+#grep --quiet -i "error:" "$capture_file" && logger "logrotate encountered an error on $(hostname): $capture_file" || true
+
+# same as above, but we're filtering out a bogus case logrotate is letting through in Ubuntu 18.04
+set pipefail
+cat "$capture_file" | python /usr/local/bin/logrotate_noise_filter.py && logger "logrotate encountered an error on $(hostname): $capture_file" || true
```
giorgiosironi: should this forced to be `python3` or it's generic?

elife/logrotate.sls
lsh-0: `scripts/` has always had much fewer files and their home isn't necessarily fixed to one location like a config file might be. It's a pedantic point and I don't mind renaming them if you find it easier

elife/config/etc-cron.daily-logrotate
lsh-0: it's a good point, I haven't tested this on python2. `python` on my machine points to `python3` but points to `python2` on Ubuntu.

lsh-0: it's generic, it works on both py2+3, but making python3 explicit

PR #252, opened 2019-07-22
Fix: postgresql-container, moves disable of host psql into own state

lsh-0:
* adds a note about it's temporariness

### DIFF
--- a/elife/postgresql-container.sls
+++ b/elife/postgresql-container.sls
@@ -26,14 +26,26 @@ docker-compose-postgres-containers.env:
         - require: 
             - docker-compose-postgres
 
+# 2019-07-22: postgresql-client.sls not present on projects where postgresql-container.sls is present (digests, profiles)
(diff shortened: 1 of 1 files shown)


lsh-0:
bit of an aside, but what if we ran highstate a second time after the first successful highstate? Just for formulas. It wouldn't take as long as the initial highstate and might reveal a class of bugs that don't present themselves until the next time around.

In this case, if you were to do a `systemctl stop postgresql` followed by the `rm -f /lib/systemd/system/postgres*`, it would work the first time but not the second. 

Just an example, I know that's not exactly what is happening here.

giorgiosironi:
> what if we ran highstate a second time after the first successful highstate? Just for formulas.

Could be an additional check to add after `pr-fresh`. It usually takes a bit less than `pt-base` and `pr-update`, which do run 2 highstates.

PR #254, opened 2019-09-03
Build php-gearman from source on 18.04

thewilkybarkid:
https://github.com/elifesciences/issues/issues/4789#issuecomment-527441303

### DIFF
--- a/elife/gearman-client-php.sls
+++ b/elife/gearman-client-php.sls
@@ -15,10 +15,15 @@ php-ppa-gearman:
         - onlyif:
             # we're using a third party ppa for php and need the separate ppa for gearman
             - test -e /etc/apt/sources.list.d/ondrej-php-trusty.list || test -e /etc/apt/sources.list.d/ondrej-ubuntu-php-xenial.list
-            
+
+gearman-php-extension:
+    pkg.installed:
+        - pkgs:
+            - php-gearman
+
 {% elif salt['grains.get']('osrelease') == '16.04' %}
 
-# the ppa in the php7 sls file references php-gearman but does not 
+# the ppa in the php7 sls file references php-gearman but does not
 # include php-gearman's dependencies:
 # * https://github.com/oerdnj/deb.sury.org/issues/711
 # * https://www.patreon.com/posts/gearman-now-in-14627464
@@ -30,7 +35,7 @@ php-ppa-gearman:
 php-ppa-gearman:
     cmd.run:
         - name: |
-            # will successfully install key but also fails (on 16.04 only) with 
+            # will successfully install key but also fails (on 16.04 only) with
             # "'ascii' codec can't decode byte 0xc5 in position 92: ordinal not in range(128)"
             # this is a Ubuntu problem, not Salt
             apt-add-repository -y ppa:ondrej/pkg-gearman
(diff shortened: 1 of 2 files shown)


elife/gearman-client-php.sls
```diff
@@ -39,9 +44,51 @@ php-ppa-gearman:
...
+        - pkgs:
+            - libgearman-dev
+    file.managed:
+        - name: /etc/php/7.2/mods-available/gearman.ini
+        - contents: extension=gearman.so
+        - require:
+            - php
+    cmd.run:
+        - cwd: /opt/pecl-gearman
+        - name: |
```
giorgiosironi: this becomes a bash script, so I think you need a `set -e` as the first line to make errors interrupt it

elife/gearman-client-php.sls
```diff
@@ -39,9 +44,51 @@ php-ppa-gearman:
...
             - php-gearman
+
+{% elif salt['grains.get']('osrelease') == '18.04' %}
+
+php-ppa-gearman:
+    pkg.purged:
+        - pkgs:
+            - php-gearman
+        - require_in:
+            - gearman-php-extension
```
giorgiosironi: IIRC for `require_in` you need to specify `- git: gearman-php-extension` with the state type or it fails silently; I almost never use it in favor of `require` from the other side

elife/gearman-client-php.sls
```diff
@@ -39,9 +44,51 @@ php-ppa-gearman:
...
+
+gearman-php-extension:
+    git.latest:
+        - name: https://github.com/wcgallego/pecl-gearman
+        - rev: gearman-2.0.5
+        - branch: master
```
giorgiosironi: since `rev` is a tag, this could possibly be unset to leave behind a detached HEAD or similar (can't remember the details)

elife/gearman-client-php.sls
lsh-0: > I almost never use it in favor of require from the other side

I don't see anything wrong with using it, but it is awfully frustrating when it fails silently because you forgot to do the fully qualified state name thing

elife/gearman-client-php.sls
lsh-0: done

elife/gearman-client-php.sls
lsh-0: fixed

elife/gearman-client-php.sls
lsh-0: yes ... I *think* in this case, because `branch` is specified as well, that revision (tag/branch) will checked out and the value for `branch` ('master') will be used locally to identify itself when you do a `git branch --show-current`. So if you're not checking out a revision from the master branch it can get misleading.

lsh-0:
I can see the new commands successfully run locally in vagrant but gearman still fails - I suspect it's because the gearman-job-server isn't being restarted. Running highstate again I suspect the same will happen as nothing is telling it it's current state has changed. I'll add a watch in on the search formula if this proves true

lsh-0:
ah - gearman-job-server service is being restarted but it happened after the search app tried and failed to `/bin/console search:setup --delete --env=dev`

highstate is working consistently now. will try a few more things before merging this and running it on ci and end2end

lsh-0:
gearman is in a failed state after restarting the instance because it can't find postgresql, presumably because it hasn't been started yet. This isn't worth investigating as the first thing that tends to happen after a restart is a highstate that brings everything in line.

PR #255, opened 2019-09-11
Postgresql 10 support + migration

lsh-0:
The idea is:

* replace `postgresql.sls` with `postgresql-10.sls` in state topfile
* run highstate
* psql-v10 is installed alongside psql-v9
* migration script checks for an unmigrated database
* if found, runs the pg_upgrade script 
* purges old psql
* replaces new psql config that is strange because it's accommodating two instances of psql
* ...? profit.


elife/config/etc-postgresql-10-main-postgresql.conf
```diff
@@ -0,0 +1,660 @@
...
+
+
+#------------------------------------------------------------------------------
+# CONFIG FILE INCLUDES
+#------------------------------------------------------------------------------
+
+# These options allow settings to be loaded from files other than the
+# default postgresql.conf.
+
+include_dir = 'conf.d'			# include files ending in '.conf' from
```
lsh-0: hrm, maybe this is what I want instead

elife/postgresql-10.sls
```diff
@@ -0,0 +1,154 @@
...
+{% endif %}
+
+
+# runs pg_upgrade on 9.4 data and then purge postgresql-9.4 
+psql-9.4 to psql-10 migration:
+    file.managed:
+        - name: /root/upgrade-postgresql-9.4-to-10.sh
+        - source: salt://elife/scripts/root-upgrade-postgresql-9.4-to-10.sh
+
+    cmd.script:
```
lsh-0: migrate even if using RDS? I think so

elife/scripts/root-upgrade-postgresql-9.4-to-10.sh
```diff
@@ -0,0 +1,108 @@
...
+        --old-datadir "/var/lib/postgresql/9.4/main" \
+        --new-datadir "/var/lib/postgresql/10/main" \
+        --old-bindir "/usr/lib/postgresql/9.4/bin" \
+        --new-bindir "/usr/lib/postgresql/10/bin" \
+        --old-options "-c config_file=/etc/postgresql/9.4/main/postgresql.conf" \
+        --new-options "-c config_file=/etc/postgresql/10/main/postgresql.conf" \
+    && {
+        # pg_upgrade succeeded
+        
+        # create a backup of the data
```
lsh-0: hrm. this would probably be a second backup if the 'copy' method is being used pg_upgrade. revisit

elife/scripts/root-upgrade-postgresql-9.4-to-10.sh
```diff
@@ -0,0 +1,108 @@
...
+        bdir=/tmp/var-lib-postgresql-9.4--backup/
+        mkdir -p "$bdir"
+        
+        # copy *and* tarball? revisit this
```
lsh-0: ha.

elife/config/etc-postgresql-10-main-postgresql.conf
giorgiosironi: adding files there rather than overwriting this one? Yes could simplify maintenance a lot

elife/postgresql-10.sls
```diff
@@ -0,0 +1,154 @@
...
+            port: 5432
+
+{% if salt['elife.cfg']('cfn.outputs.RDSHost') %}
+pgpass-rds-entry:
+    file.append:
+        - name: /root/.pgpass
+        - source: salt://elife/config/root-pgpass
+        - template: jinja
+        - defaults:
+            user: {{ pillar.elife.db_root.username }}
```
giorgiosironi: shouldn't this be `salt['elife.cfg']('project.rds_username')`? RDS provisioning by builder surely can't know the value of a pillar, even if they will probably coincide (`root` being the value?)

elife/postgresql-10.sls
giorgiosironi: Depends what the script does. But if you need to change the RDS version you should go through `update_infrastructure` rather than relying on Salt. In general, we can't have a stack with RDS but without EC2 for this reason (Salt doesn't run to complete the setup) so would avoid the coupling.

elife/postgresql-10.sls
```diff
@@ -0,0 +1,154 @@
...
+        - template: jinja
+        - mode: 0600
+        - defaults:
+            user: {{ pillar.elife.db_root.username }}
+            pass: {{ pillar.elife.db_root.password }}
+            host: localhost
+            port: 5432
+
+{% if salt['elife.cfg']('cfn.outputs.RDSHost') %}
+pgpass-rds-entry:
```
giorgiosironi: shouldn't this be the _only_ thing we do is RDS is present? We shouldn't need to install any package or setup any service as `postgresql-client.sls` should provide `psql`.

I may not be fully aware of use cases where RDS needs to run alongside a local `postgres` daemon though.

elife/postgresql-10.sls
giorgiosironi: This script relies on `systemctl` and `/var/lib/postgresql/`, can't work with RDS anyway?

lsh-0: sorry, I was unclear. Upgrading RDS will happen through `update_infrastructure`, I was just wondering if psql 9.4 data should be upgraded to psql 10 even if local psql isn't being used. I think for consistency it should be. It's just an empty database most likely.

elife/postgresql-10.sls
lsh-0: we have one case that I've hated since forever: elife-dashboard (rds) + article-scheduler (local psql).

elife/postgresql-10.sls
lsh-0: I'll take a closer look

lsh-0: we seem to have hardcoded the rds username into the CloudFormation template: https://github.com/elifesciences/builder/blob/master/src/buildercore/cfngen.py#L312

and `db_root` is also using the same value: 
https://github.com/elifesciences/builder-private/blob/master/pillar/elife.sls#L191-L198

while we can rely on this value being constant, its the rds password that needs to be pulled from the buildvars

lsh-0: also, it's not something I want to be changing in this PR. `postgresql-10.sls` is pretty much a copy+paste of `postgresql.sls` with a few things tweaked.

lsh-0: marking this as resolved. Let me know if you want me to address it further

elife/postgresql-10.sls
giorgiosironi: Agreed

elife/postgresql-10.sls
giorgiosironi: I hate that one, too.

elife/postgresql-10.sls
giorgiosironi: Agree this all works in this porting. Would address separately as the pillar can't really be changed at the moment without this breaking, I guess.

elife/scripts/root-upgrade-postgresql-9.4-to-10.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
```
giorgiosironi: have never done this, so can't provide much feedback unfortunately

lsh-0: I found some evidence I may have done it with psql 8 to 9, but I don't remember having done it either to be honest.

lsh-0:
fyi, the new version of Postgresql has changed to **11** from **10** - this is because RDS won't let us upgrade 9.4 to 10. I'm sure there is a good reason, I just can't find it except for this: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html#USER_UpgradeDBInstance.PostgreSQL.MajorVersion.Multi

I've just finished testing the changes to the migration script on a couple of local vagrant projects with no difference in behaviour. The uwsgi app failed to come up but that was project related, more here: https://github.com/elifesciences/issues/issues/5111

PR #257, opened 2019-09-19
Pandoc

gnott:
I wanted to see if I could find a way to allow `elife-bot-formula` to include `pandoc` on an instance. I need at least version 2.5 so the default Ubuntu release package is not new enough.

Here's my guess at downloading the 64 bit Debian package and installing the executable. I'm not sure how the permissions will get set, or if this will even run. It's a concept to be considered?

Pandoc installation page https://pandoc.org/installing.html#linux

This is using the same version of the package I use for travis-ci tests in the https://github.com/elifesciences/decision-letter-parser library.

### DIFF
--- a/elife/pandoc.sls
+++ b/elife/pandoc.sls
@@ -1,0 +1,8 @@
+pandoc:
+    file.managed:
+        - name: /root/pandoc-2.5-1-amd64.deb
+        - source: https://github.com/jgm/pandoc/releases/download/2.5/pandoc-2.5-1-amd64.deb
+
+    cmd.run:
+        - name: sudo dpkg -i pandoc-2.5-1-amd64.deb
+        - cwd: /root


lsh-0:
I ran in to this problem with pandoc as well while experimenting with converting JATS to HTML. Even 18.04 LTS didn't have the feature I needed at the time and I had to use 18.10 to get what I was after:

https://github.com/elifesciences/jats-xml-to-pdf/blob/master/pandoc/Dockerfile

that project was using containers. elife-bot has container support available and it's being used with strip-coverletter. If you can, I'd suggest using a container for this.

Otherwise, check that `.deb` is built for and compatible with Ubuntu 16.04 (what elife-bot uses).

Next, the `cmd.run` state is only used when you absolutely have to use it. Use [`pkg.installed`](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkg.html#salt.states.pkg.installed) instead:

```yaml
mypkgs:
  pkg.installed:
    - sources:
      - foo: salt://rpms/foo.rpm
      - bar: http://somesite.org/bar.rpm
      - baz: ftp://someothersite.org/baz.rpm
      - qux: /minion/path/to/qux.rpm
```

gnott:
Thanks for the additional example. I am creating a python library that would be installed via pip, and I cannot seem to wrap my head around why a python module should require Docker, however it would sort of require pandoc, so I guess it's the same thing. I have a couple ways of possibly invoking this one pandoc call so I'll explore those when I get further along with the library's core logic.

lsh-0:
> Thanks for the additional example. I am creating a python library that would be installed via pip, and I cannot seem to wrap my head around why a python module should require Docker, 

well, pandoc isn't a python module, it's haskell iirc, so if your library is making external calls to system programs, and the specific version of the program you want isn't available or compatible with the one you're current running, then one option is to run that in a container. Making system calls to a container is no different or stranger than calling any other external program, like imagemagick.

If the version of pandoc you require is available in 18.04, I'd suggest we upgrade elife-bot to 18.04 (which is going to happen eventually anyway).

giorgiosironi:
I second the container or upgrade to 18.04 routes. Isolated `.deb` packages can be a pain because of the lack of a way to fetch dependencies, though they are helpful for self-contained software that you want to update often like Jenkins.

gnott:
The details and reasons why are helpful to me, thanks.

I think the earliest potential Ubuntu release with a sufficient `pandoc` version would be `18.10`, so it might be a while before a native `pandoc` executable is available.

For now, I think I'm going to allow the particular parser library to try `pandoc` first, and if that fails it will try Docker. If both those fails, then using the Python library would be limited to the formatting and output it is able to produce, without gathering input from a `.docx` file.

Interesting, not unexpected, observations is when I run the test case in the project `pandoc` can finish them in `3.35` seconds, the docker method in `24.90` seconds.

I'll close this PR considering the decision here and the future may bring an executable one day!

PR #258, opened 2019-10-08
Feature: adds a 60 minute timeout to highstate call in daily update script

lsh-0:


lsh-0:
send a further KILL command if process still running after initial TERM command.

The failure will be captured and reported by loggly, bu we could also capture the exit code of `timeout` to log a different error.

lsh-0:
me too, merging.

giorgiosironi:
Temporarily reverting in https://github.com/elifesciences/builder-base-formula/commit/e03ca3ff67c8537675ce2c946d70cfe3a7a848dd because https://alfred.elifesciences.org/job/process/job/process-daily-updates-prod started to hang.

PR #259, opened 2019-10-11
Feature: docker container version

lsh-0:
- [x] depends on https://github.com/elifesciences/builder-configuration/pull/88


elife/config/home-deploy-user-postgres-.env
```diff
@@ -1,2 +1,2 @@
 COMPOSE_PROJECT_NAME=postgres
-IMAGE_TAG=9.4.16
+IMAGE_TAG={{ pillar.elife.docker_postgresql.image_version }}
```
giorgiosironi: it is more correct to say this is an `image_tag` rather than an `image_version`, because you could use e.g `11` or `11-alpine` or `11.5-alpine`, etc

lsh-0: done

PR #261, opened 2019-10-14
removes postgresql-client.sls

lsh-0:
states have been merged into postgres-11.sls and postgres-container.sls


lsh-0:
giving this one a kick

PR #263, opened 2019-10-15
Reintroduce timeout with --foreground

giorgiosironi:
For https://github.com/elifesciences/issues/issues/5163

According to https://www.gnu.org/software/coreutils/manual/html_node/timeout-invocation.html this seems to imply that the combination of `docker-compose exec` and `sudo` is "interactive" e.g. reads from the terminal.


lsh-0:
looks fine :+1:

PR #265, opened 2019-12-13
salt upgrade, these get basebox building without warnings or errors

lsh-0:


elife/hostname.sls
```diff
@@ -15,7 +15,9 @@ set-hostname:
 
 set-hosts:
     host.present:
-        - ip: 127.0.0.1
+        - ip: 
+            - 127.0.0.1
+            - ::1
```
lsh-0: state updated in Salt `2018.3`, now issues a warning if name is present for other IPs not in list

* https://docs.saltstack.com/en/latest/ref/states/all/salt.states.host.html#salt.states.host.present

lsh-0:
ah - it's still using 14.04 in the project config. I have another PR for that somewhere ...

PR #268, opened 2019-12-17
python2 removal / salt upgrade

lsh-0:
* adds notes against base packages with py2 requirements
* instances of unqualified `virtualenv` and `pip` usage addressed


elife/mysql-client.sls
```diff
@@ -4,5 +4,4 @@ mysql-package-list:
     pkg.installed:
         - pkgs:
             - mysql-client
-            - python-mysqldb # py2, to be removed
```
lsh-0: turns out we need to keep this until we're running on upgraded Salt. Makes sense

elife/aws-cli.sls
```diff
@@ -1,4 +1,3 @@
 aws-cli:
     cmd.run:
-        - name: pip install --upgrade 'awscli~=1.16'
-
+        - name: python3 -m pip install --upgrade 'awscli~=1.16'
```
giorgiosironi: good to know another way to invoke pip from Python 3. I've been using `pip3`so far, which I think is equivalent

elife/base.sls
```diff
@@ -24,8 +22,12 @@ base:
             # provides add-apt-repository binary needed to install a new ppa easily
             # renamed in 18.04
             {% if osrelease in ['14.04', '16.04'] %}
```
giorgiosironi: remove `14.04` from the list?

elife/base.sls
```diff
@@ -24,8 +22,12 @@ base:
...
+            # depends on py2
```
giorgiosironi: so can't do anything about this one, right?

elife/nodejs6.sls
```diff
@@ -3,10 +3,6 @@ nodejs6:
     pkgrepo.managed:
         - name: deb http://deb.nodesource.com/node_6.x {{ distro }} main
```
giorgiosironi: we really need to move off this :scream_cat:

elife/aws-cli.sls
lsh-0: yup.

I prefer the longish `python3 -m foo bar` approach because then you only need to figure out which python3 is being used. `pip` and `pip3` within a virtualenv are identical if it's a python3 venv, `pip` and `pip3` outside a virtualenv are not.

elife/base.sls
lsh-0: right, not until we get the dozen projects on 16.04 on to 18.04: https://github.com/elifesciences/issues/issues/5095

even then ... well, cross that bridge when we get there

elife/nodejs6.sls
lsh-0: one overwhelming thing at a time

lsh-0:
I don't work for eLife Wednesdays instead of Thursdays starting in 2020, so I'll wait until Thurs to merge this in so I can deal with any fallout Thurs+Fri.

elife/nodejs6.sls
giorgiosironi: it's an application problem, rather than an infrastructure one

PR #270, opened 2020-01-03
Switch proofreader Vagrant default to 0.2

giorgiosironi:
Matches
https://github.com/elifesciences/builder-private/blob/3b6cb4539d13a2b42165cff93f04345c00389c3f/pillar/elife.sls#L300


giorgiosironi:
Will merge on green, only affects running static checks within Vagrant.

PR #271, opened 2020-01-21
/usr/local/utils

lsh-0:
* moves contents of /usr/local/utils to /usr/local/bin
* adds a comment to improve grep'ability of the scripts being copied.

### DIFF
--- a/elife/base.sls
+++ b/elife/base.sls
@@ -37,6 +37,9 @@ base:
             - sysstat # provides iostat
             - iotop
 
+            # useful for smoke testing the JSON output
+            - jq
+
 base-purging:
     pkg.purged:
         - pkgs:
--- a/elife/utils.sls
+++ b/elife/utils.sls
@@ -15,4 +15,3 @@ utils-scripts:
 utils-scripts-path:
     file.absent:
         - name: /etc/profile.d/utils-path.sh
-        - source: salt://elife/config/etc-profile.d-utils-path.sh
--- a/elife/utils.sls
+++ b/elife/utils.sls
@@ -1,25 +1,18 @@
 utils-scripts:
     file.recurse:
-        - name: /usr/local/utils/
+        # this state copies the scripts:
+        # build_vars, retry, set_local_revision, update_python_dependency, wait_for_port
+        # this directory was only available in interactive login shells
+        # it was not available to commands executed using su or sudo
+        # for example, running a cmd.run state with a user other than root
+        # or sudo'ing to root to run a command
+        #- name: /usr/local/utils/
+        # this directory is always on the PATH
+        - name: /usr/local/bin
         - source: salt://elife/utils
         - file_mode: 555
 
 utils-scripts-path:
-    file.managed:
+    file.absent:
         - name: /etc/profile.d/utils-path.sh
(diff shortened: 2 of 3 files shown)


lsh-0:
this state is running and failing but reporting success every time in Vagrant:
https://github.com/elifesciences/iiif-formula/blob/master/salt/iiif/loris.sls#L278-L284

on `iiif--prod--3` (and the other nodes probably) this is succeeding somehow:

```
[INFO    ] Running state [wait_for_port 80
loris-smoke] at time 04:14:20.472515
[INFO    ] Executing state cmd.run for [wait_for_port 80
loris-smoke]
[INFO    ] Executing command 'wait_for_port 80
loris-smoke' as user 'elife' in directory '/home/elife'
[INFO    ] {'pid': 8680, 'retcode': 0, 'stderr': '', 'stdout': 'Waiting for someone listening on port 80\nConnection was established on port 80\n> http://localhost/\n    [ OK ] 2xx Response code\n> http://localhost/ping\n    [ OK ] 2xx Response code\nOK (2/2)'}
[INFO    ] Completed state [wait_for_port 80
loris-smoke] at time 04:14:20.572729 duration_in_ms=100.214
```
Salt does a `su` to the user `elife` to run the command. Here is my result of the same:
```
elife@prod--iiif:~$ whereis wait_for_port
wait_for_port: /usr/local/utils/wait_for_port
elife@prod--iiif:~$ echo $PATH
/usr/local/utils:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
elife@prod--iiif:~$ sudo su
root@prod--iiif:/home/elife# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
root@prod--iiif:/home/elife# su elife
elife@prod--iiif:~$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
```

my guess is that Salt is explicitly preserving certain environment variables (like PATH) when doing the `su`, like specified in `sudoers` but some thing has changed between 2017.7 (remote) and 2018.3 (vagrant) or there is something about my local dev environment being carried into Vagrant ... no idea at all.

but this PR should remove any magic

elife/base.sls
```diff
@@ -37,6 +37,9 @@ base:
             - sysstat # provides iostat
             - iotop
 
+            # useful for smoke testing the JSON output
+            - jq
```
giorgiosironi: I install this through the OS packages so I think the version will be alright

elife/utils.sls
```diff
@@ -1,25 +1,18 @@
...
+        # this directory is always on the PATH
+        - name: /usr/local/bin
         - source: salt://elife/utils
         - file_mode: 555
 
 utils-scripts-path:
-    file.managed:
+    file.absent:
         - name: /etc/profile.d/utils-path.sh
         - source: salt://elife/config/etc-profile.d-utils-path.sh
```
giorgiosironi: don't need `source`?

lsh-0: you're right, we don't. I couldn't rm that file as well

PR #272, opened 2020-01-30
Remove unused Node.js installations

giorgiosironi:
Compare https://github.com/elifesciences/builder-configuration/blob/master/salt/top.sls where only `elife.nodejs6` is used. That needs upgrading, but not to any of these as they are also not supported anymore.


lsh-0:
We also need to remove the apache and telegraf states and probably a bunch more

PR #273, opened 2020-02-03
patches python3-mysqldb 'cursors.py', decoding a bytes warning to a str

lsh-0:


elife/mysql-client.sls
```diff
@@ -6,3 +6,16 @@ mysql-package-list:
...
+
+    # on Ubuntu 16.04, in Salt 2018.3, using Python3, there is a problem passing a 
+    # bytestring as an argument to a function that expects a regular string.
+    # this patch decodes any bytestring arguments into a regular string
+    file.patch:
+        - name: /usr/lib/python3/dist-packages/MySQLdb/cursors.py
+        - source: salt://elife/config/usr-lib-python3-dist-packages-MySQLdb-cursors.py.patch
+        # md5 of original file: aca2fb480a76e30f65601018916e86bd
+        # md5 of patched file:
+        - hash: 15d883d28d4548374522fad16eed560e
```
giorgiosironi: isn't this likely to change if the file is changed upstream in a way that is still compatible with the patch?

elife/mysql-client.sls
```diff
@@ -6,3 +6,16 @@ mysql-package-list:
             - mysql-client
             - python-mysqldb # py2, Salt 2017.7.x
             - python3-mysqldb # py3, Salt 2018.3+
+
+    # on Ubuntu 16.04, in Salt 2018.3, using Python3, there is a problem passing a 
+    # bytestring as an argument to a function that expects a regular string.
+    # this patch decodes any bytestring arguments into a regular string
+    file.patch:
+        - name: /usr/lib/python3/dist-packages/MySQLdb/cursors.py
+        - source: salt://elife/config/usr-lib-python3-dist-packages-MySQLdb-cursors.py.patch
```
giorgiosironi: where does this come from? If it has been downloaded needs a URL, if handwritten state it

giorgiosironi:
Patching failing on the `pr-fresh` variant now, seems to be at the `patch` command level rather than in the MD5 check.

elife/mysql-client.sls
lsh-0: the patch? I made the patch. I'll make a note of it's origin

elife/mysql-client.sls
lsh-0: Then the hash check will fail and we'll know it's been fixed or the patch needs updating

It's less likely the `python3-mysqldb` package will change in 16.04 than the Salt module code.

I don't like patching these files but I don't see any movement in the ticket I opened with the Salt guys.

PR #274, opened 2020-02-04
Remove 14.04 conditional PHP branch

giorgiosironi:
Since to my understanding we don't run that OS anymore, simplifying.


lsh-0:
I have a ticket somewhere to go through the formulas and remove all the conditional stuff for past supported OSs but feel free to remove any as you encounter them.

PR #276, opened 2020-02-10
/src/elife/spectrum-app.cfg, removes all mentions of peerscout

lsh-0:
fyi @giorgiosironi


giorgiosironi:
I suspect https://github.com/elifesciences/elife-spectrum/blob/master/spectrum/checks.py#L1085-L1089 is now broken though.

lsh-0:
sorry about that, the change did feel like a stab in the dark. If peerscout is gone then that check should disappear as well

PR #277, opened 2020-02-10
Remove 14.04 reference from OpenJDK installation

giorgiosironi:
/cc @lsh-0


giorgiosironi:
Not sure if blocked or not. The PPA is still available but now should be considered a risky dependency, see https://launchpad.net/~jonathonf/

PR #278, opened 2020-02-19
Deploy New Relic PHP configuration file on PHP 7.2

giorgiosironi:
Appears this is not deployed, leading to applications reporting with some default values such as `PHP Application` rather than their name. Reporting still works because the installation process somehow puts the right license key in the default `newrelic.ini` file, which is used in place of this.

### DIFF
--- a/elife/newrelic-php.sls
+++ b/elife/newrelic-php.sls
@@ -43,7 +43,7 @@ newrelic-install-script:
             #- newrelic-php-extension-headless-configuration
 
 
-{% if osrelease == "18.04" %}
+{% if salt['grains.get']('osrelease') == "18.04" %}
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Definition of done is that https://rpm.newrelic.com/accounts/1451451/applications/232979756 (`PHP Application`) should become empty. Mostly I see `search--prod` processes there, but there may be something else.

giorgiosironi:
Running also https://alfred.elifesciences.org/job/prod-personalised-covers/ because I see `personalised-covers` too.

giorgiosironi:
https://rpm.newrelic.com/accounts/1451451/applications/232979756 is empty now.

PR #284, opened 2020-03-10
python, bounds virtualenv version to avoid salt bug

lsh-0:


elife/python.sls
```diff
@@ -35,7 +35,9 @@ python-dev:
 
 global-python-requisites:
     cmd.run:
-        - name: python3 -m pip install "virtualenv>=13"
```
giorgiosironi: I imagine the constraint on 13 as a minimum is not necessary anymore?

lsh-0: Almost definitely not. It would have been put there to force an upgrade at a guess. And it's probably been there since 14.04 days at least.

PR #286, opened 2020-04-07
Remove unused elife.php-dummies

giorgiosironi:
Nothing uses this anymore, in favor of [`elife.sidecars`](https://github.com/elifesciences/builder-configuration/blob/master/salt/top.sls#L378) probably which is not language-specific.


giorgiosironi:
[Pillars](https://github.com/search?q=org%3Aelifesciences+php_dummies&type=Code) can be removed as a follow-up.

PR #287, opened 2020-04-15
sidecars, healthcheck fix

lsh-0:
this removes the `healthcheck: None` we're seeing in `profiles`.

only personalised covers is using the health check:

* https://github.com/elifesciences/builder-configuration/blob/4b085410a4c9923613f97c273e1a75b975a812b9/pillar/personalised-covers-ci-public.sls#L16
* https://github.com/elifesciences/personalised-covers-formula/blob/6c12a674b8e5f50e2f014340088d42d8007a3ef9/salt/pillar/personalised-covers.sls#L34

### DIFF
--- a/elife/sidecars.sls
+++ b/elife/sidecars.sls
@@ -23,7 +23,7 @@ docker-compose-{{ configuration['name'] }}:
             ports: {{ configuration.get('ports', {})|yaml }}
             environment: {{ configuration.get('environment', {})|yaml }}
             volumes: {{ configuration.get('volumes', [])|yaml }}
-            healthcheck: {{ configuration.get('healthcheck') }}
+            healthcheck: {{ configuration.get('healthcheck')|yaml }}
         - makedirs: True
         - require: 
             - deploy-user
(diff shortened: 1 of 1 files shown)


lsh-0:
yes, this will bork personalised covers:

```
[CRITICAL] Rendering SLS 'base:elife.sidecars' failed: while parsing a block mapping
  in "<string>", line 1, column 1:
    docker-network-sidecars:
    ^
expected <block end>, but found '<scalar>'
  in "<string>", line 27, column 3:
      s3 ls'
      ^
local:
    Data failed to compile:
----------
    Rendering SLS 'base:elife.sidecars' failed: while parsing a block mapping
  in "<string>", line 1, column 1:
    docker-network-sidecars:
    ^
expected <block end>, but found '<scalar>'
  in "<string>", line 27, column 3:
      s3 ls'
      ^
```

I was thinking last night it wasn't the correct solution. At least I have some insight into roughly what is happening now.

lsh-0:
alright, much better now. For some reason `None` is being interpreted as the string `"None"` which is truth-y, so the healthcheck section is printed out with a `healthcheck: None`. However an empty string is still false-y in Python and using that as a default elides the non-existent healthcheck as intended.

PR #306, opened 2021-05-25
composer, bumps composer from 1.10.16 to 1.10.21

lsh-0:
New github tokens break this version of composer. 

Doesn't present in dev environments without a token.


lsh-0:
base is broken after github token update, merging

PR #307, opened 2021-06-29
composer, pulls composer version from builder-configuration

lsh-0:


elife/composer.sls
```diff
@@ -80,15 +78,14 @@ composer-global-paths:
         - require:
             - file: composer-home-dir
 
-# lsh@2021-05-25: requisites would seem to preclude this from running:
-# 'run before composer is installed' and 'only if composer is installed'
-# todo: fix or remove
 update-composer:
```
lsh-0, line 81: might even be able to get rid of this state altogether - specifying `--version` in install-composer seems to be enough

PR #311, opened 2021-07-22
python.sls, pins version of virtualenv installed.

lsh-0:


lsh-0:
fyi @NuclearRedeye , cause of the failed highstate on elife-alfred--prod overnight. I fixed the machine but highstate will continue to fail until either the virtualenv bug is fixed or this fix goes in.

PR #312, opened 2021-08-16
feat: add initial script for creating era jobs on bot message queues

NuclearRedeye:


NuclearRedeye:
@giorgiosironi could I ask for some help understanding how to get the `pr-base` stage back to green? You can see from the history of this PR that both the `pr-base` and `pr-fresh` were originally failing when trying to install `helm`. You can see from the snippet of log below that the package location has changed. I updated the formula to point at a new location for helm releases and that change got the `pr-fresh` stage green, but the `pr-base` stage is still using the older formula without my fix.

```
[2021-08-17T07:45:47.968Z] [34.227.223.254] out: [ERROR   ] Failed to cache https://storage.googleapis.com/kubernetes-helm/helm-v2.11.0-linux-amd64.tar.gz: Error: HTTP 403: Forbidden reading https://storage.googleapis.com/kubernetes-helm/helm-v2.11.0-linux-amd64.tar.gz
```

NuclearRedeye:
Bringing this to your attention @lsh-0, didn't hear back from Giorgio and haven't had a chance to chase him via another medium yet. See my previous comment, not sure if you have any insight at all? If not, don't worry about it.

lsh-0:
I've opened a separate PR for this problem: https://github.com/elifesciences/builder-base-formula/pull/314

(same solution)

> could I ask for some help understanding how to get the pr-base stage back to green? 

Because these formulas depend on the outside world so much it's possible for them to break without them ever changing.

In this case, a URL we were using became inaccessible and broke the formula.

`pr-fresh` is green because it builds the updated formula from scratch with the change included. `pr-head` and `pr-base` build the formula from known states in the formula's history, then applies the change and runs highstate again. 

However, from it's current state those two builds won't ever pass unless that URL fixes itself (which it won't). What we do when this happens is declare the reason why the formula isn't passing, ensure the `fresh` build is green and merge.

Some methods to avoid this is keeping a mirror of every single dependency a project requires. All sorts of benefits with this approach but difficult to pull off without developers pulling a mutiny ;)

PR #314, opened 2021-08-23
helm.sls, fixes location of helm script

lsh-0:


lsh-0:
an S3 bucket we rely on is broken. pr-head and pr-base won't ever be green.

PR #321, opened 2021-11-27
Install poppler-utils

gnott:
Re issue https://github.com/elifesciences/issues/issues/7044
Also to issue https://github.com/elifesciences/issues/issues/7029

How to make the executables included in the package https://launchpad.net/ubuntu/bionic/+package/poppler-utils hasn't been discussed in full, probably, but I thought I would take a shot at how it might be installed as part of the instance OS.

Alernatively, we may prefer to use Docker to invoke them.

Please feel welcome to add more commits to this PR, and I may have this wrong, nor did I test the `apt-get install` command on an Ubuntu system.

### DIFF
--- a/elife/poppler-utils.sls
+++ b/elife/poppler-utils.sls
@@ -1,0 +1,7 @@
+# install Ubuntu PDF utilities package 
+
+poppler-utils:
+    cmd.run:
+        - name: apt-get install poppler-utils -y
+        - unless:
+            - pdfinfo


lsh-0:
Is it just elife-bot that will need these? If so, I'll add them to the elife-bot formula. When we have ~3 cases of formulas doing the same thing I tend to add it to builder-base at that point. 

In the beginning a *lot* of stuff went into builder-base, and I'm trying to [prune some of that away now](https://github.com/elifesciences/builder-base-formula/pull/320).

 With Salt, I tend to use `cmd.run` as a last resort. Installing packages is usually done with:

```yaml
poppler-utils:
    pkg.installed:
        - name: poppler-utils
```

which would handle your `unless` clause automatically. For installing many packages at once you can do this:

```yaml
poppler-utils:
    pkg.installed:
        - pkgs:
            - poppler-utils
            - ...
```

`pkg.installed`, `file.managed` and `git.present` are what we tend to use the most.

Index of Salt states here: https://docs.saltproject.io/en/latest/ref/states/all/index.html

lsh-0:
see here: https://github.com/elifesciences/elife-bot-formula/pull/84

PR #337, opened 2022-03-28
base, adds rg and fd to 20.04 instances

lsh-0:


elife/base.sls
```diff
@@ -59,6 +58,13 @@ base-purging:
         - require:
             - base
 
+symlink fdfind to fd:
```
lsh-0, line 61: this is 20.04 only

PR #426, opened 2024-10-08
upgrade builder base formula to support psql13 elifesciences/issues#8969

ali-tm-amin:


salt/elife/postgresql-13.sls
```diff
@@ -0,0 +1,162 @@
...
+            - pgpass-file
+{% else %}
+    service.dead:
+        - enable: False
+        - require:
+            - pkg: postgresql
+            - pgpass-file
+{% endif %}
+
+{% if not salt['elife.cfg']('cfn.outputs.RDSHost') %}
```
scottaubrey, line 64: I'm guessing this file no longer exists, so should we just drop this state?

scottaubrey (approved):
Looks OK to me.

PR #438, opened 2025-01-17
introduce a PPA-based php state with version selector via pillar data elifesciences/issues#9141

scottaubrey:
elifesciences/issues#9141


scottaubrey:
Need to test upgrade path with an existing, non-PPA's instance

